PR #4, opened 2023-08-11
rbac changes for enhanced container insights

chadpatel:
*Issue #, if available:*

Enhanced container insights requires additional RBAC permissions beyond what container insights requires


mitali-salvi (approved):

config/rbac/agent_role.yaml
```diff
@@ -23,4 +23,7 @@ rules:
     verbs: ["create", "get"]
   - apiGroups: [""]
     resources: ["configmaps"]
-    verbs: ["update"]
\ No newline at end of file
+    resourceNames: ["cwagent-clusterleader"]
```
sky333999, line 26: We are restricting it now to just this config map. Was any other code relying on any other config maps?

chadpatel: I am not opinionated on locking it down to a specific mag.  I don't think we have code relying on any other maps.  If we were to ever add a second config map (second leader??) then this might be annoying

PR #7, opened 2023-08-29
Add unit tests for apis

lisguo:
*Description of changes:*
Adds unit tests for the amazoncloudwatchagent webhook and instrumentation webhook.

### DIFF
--- a/apis/v1alpha1/amazoncloudwatchagent_webhook_test.go
+++ b/apis/v1alpha1/amazoncloudwatchagent_webhook_test.go
@@ -1,0 +1,244 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1alpha1
+
+import (
+	"fmt"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+func TestAgentDefaultingWebhook(t *testing.T) {
+	one := int32(1)
+	five := int32(5)
+
+	tests := []struct {
+		name     string
+		agent    AmazonCloudWatchAgent
+		expected AmazonCloudWatchAgent
+	}{
+		{
+			name:  "all fields default",
+			agent: AmazonCloudWatchAgent{},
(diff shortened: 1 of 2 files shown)


apis/v1alpha1/amazoncloudwatchagent_webhook_test.go
```diff
@@ -0,0 +1,244 @@
...
+					Replicas:        &three,
+					UpgradeStrategy: "adhoc",
+					Config: `"agent": {
+   "metrics_collection_interval": 60,
+   "region": "us-west-1",
+   "logfile": "/opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log",
+   "debug": false,
+   "run_as_user": "cwagent"
+}
+`,
```
dchappa, line 129: nit - the indenting seems off here.

dchappa (approved):

apis/v1alpha1/amazoncloudwatchagent_webhook_test.go
lisguo: This is pretty much taken from https://github.com/open-telemetry/opentelemetry-operator/blob/main/apis/v1alpha1/opentelemetrycollector_webhook_test.go#L174

I think the linter was removed for the config string since yaml indentation matters. But since we use json maybe we don't need to remove the linter. Good catch!

apis/v1alpha1/amazoncloudwatchagent_webhook_test.go
```diff
@@ -0,0 +1,244 @@
...
+			expectedErr: "the AmazonCloudWatchAgent Spec Ports configuration is incorrect",
+		},
+		{
+			name: "invalid port num",
+			otelcol: AmazonCloudWatchAgent{
+				Spec: AmazonCloudWatchAgentSpec{
+					Ports: []v1.ServicePort{
+						{
+							Name: "aaaabbbbccccddd", // len: 15
+							// no port set means it's 0, which is invalid
```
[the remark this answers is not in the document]
lisguo: We do set default ports. See https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/pkg/collector/common.go#L10

But the AmazonCloudWatchAgent resource spec also lets you override the ports which is what this test is doing.

PR #11, opened 2023-09-15
Integ test beta

nathalapooja:
*Description of changes:*
Implemented the Integration testing for EKS addon on Beta cluster

Successful test run https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/6201567912


sky333999:
I had assumed all the CFN stuff was temporary to validate that it works before we align it with our existing terraform setup :)

Lets sync up offline - Id like to better understand why the beta and non-beta tests need to be so diff. Ideally, id expect we can reuse most of the cost with a flag to add some extra beta components.

sky333999:
Also some additional enhancements which can be a follow up:
* Would be good to have a way to trigger both beta and non-beta independently and via the GHAs UI.
* We need a way of taking in a version as an option arg that can be supplied via the UI.
* As part of the matrix to run multiple EKS versions (comment from last PR), we should also add ability to test diff addon versions.

nathalapooja:
> Also some additional enhancements which can be a follow up:
> 
> * Would be good to have a way to trigger both beta and non-beta independently and via the GHAs UI.
> * We need a way of taking in a version as an option arg that can be supplied via the UI.
> * As part of the matrix to run multiple EKS versions (comment from last PR), we should also add ability to test diff addon versions.

I believe we should better create a sim to track all these additional enhancements and track it.

nathalapooja:
https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/6223907660 successful integration test after the terraform beta changes.

integration-tests/eks-addon/validateResources_test.go
```diff
@@ -18,89 +18,86 @@ import (
 	"testing"
 )
 
-const NAMESPACE = "amazon-cloudwatch"
+const nameSpace = "amazon-cloudwatch"
```
sky333999, line 21: nit: nameSpace -> namespace

sky333999 (approved):

lisguo (approved):

PR #15, opened 2023-09-27
Re-enable build and upload step on push to main

lisguo:
*Issue #, if available:*
N/A

*Description of changes:*
Re-enable the build and upload github action

### DIFF
--- a/.github/workflows/build-and-upload.yml
+++ b/.github/workflows/build-and-upload.yml
@@ -1,11 +1,11 @@
 name: Build And Upload
 
-#on:
-#  push:
-#    branches:
-#      - main*
-#  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


mitali-salvi (approved):

.github/workflows/build-and-upload.yml
```diff
@@ -1,11 +1,11 @@
...
-#on:
-#  push:
-#    branches:
-#      - main*
-#  workflow_dispatch:
-#  workflow_call:
+on:
+  push:
+    branches:
+      - main*
```
sky333999, line 6: This should ideally be only main.
The way it currently is, another main-* branch could overwrite the image published from main and it would be hard to track who's publishing what.

PR #17, opened 2023-10-06
Create upload release workflow

lisguo:
*Description of changes:*
Runs the integration tests and on success, pushes the staging artifact to release.

*Tests*
Ran workflow on push to branch and saw a successful run. Used a test tag
https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/6433487066/job/17470648145

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1,3 +1,6 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
 name: Run Integration Tests
 env:
   TERRAFORM_AWS_ASSUME_ROLE: ${{ secrets.TERRAFORM_AWS_ASSUME_ROLE }}
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -13,6 +13,7 @@ on:
       - '.github/**'
       - '!.github/workflows/integration-test.yml'
   workflow_dispatch:
+  workflow_call:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
--- a/.github/workflows/upload-release.yml
+++ b/.github/workflows/upload-release.yml
@@ -1,0 +1,49 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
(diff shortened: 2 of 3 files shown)


.github/workflows/integration-test.yml
```diff
@@ -13,6 +13,7 @@ on:
       - '.github/**'
       - '!.github/workflows/integration-test.yml'
   workflow_dispatch:
+  workflow_call:
```
jefchien, line 16: What are we using workflow_call for?

.github/workflows/upload-release.yml
```diff
@@ -0,0 +1,46 @@
...
+
+name: Upload Release Artifact
+env:
+  # Use terraform assume role for uploading to ecr
+  AWS_ASSUME_ROLE: ${{ secrets.TERRAFORM_AWS_ASSUME_ROLE }}
+  ECR_OPERATOR_STAGING_IMAGE: ${{ secrets.ECR_OPERATOR_STAGING_IMAGE }}
+  ECR_OPERATOR_RELEASE_IMAGE: ${{ secrets.ECR_OPERATOR_RELEASE_IMAGE }}
+
+on:
+  workflow_dispatch:
```
jefchien, line 12: Are we going to tag the releases? Could this be triggered on release?

.github/workflows/upload-release.yml
```diff
@@ -0,0 +1,46 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
jefchien, line 1: Do we need to add the copyright header in the other workflows as well?

jefchien (approved):

.github/workflows/integration-test.yml
lisguo: this allows the upload-release workflow to call the integ test as a pre-step to upload

.github/workflows/upload-release.yml
lisguo: I have an action to come up with a versioning strategy, but for now it is just tagged as latest which is part of the secret

.github/workflows/upload-release.yml
lisguo: yeah good point, actually this header should be referencing apache

.github/workflows/integration-test.yml
jefchien: Oh right. Reusable workflows.

PR #28, opened 2023-10-30
Update configmap and service names for cloudwatch-agent

sky333999:
*Issue #, if available:*

*Description of changes:* We are trying to be consistent in our naming of our components to what our existing getting started docs refer to them as.
Accordingly, updating the default config map name and the agent service name used for default instrumentation.

### DIFF
--- a/pkg/naming/main.go
+++ b/pkg/naming/main.go
@@ -9,8 +9,8 @@ import (
 )
 
 // ConfigMap builds the name for the config map used in the AmazonCloudWatchAgent containers.
-func ConfigMap(agent v1alpha1.AmazonCloudWatchAgent) string {
-	return DNSName(Truncate("%s", 63, agent.Name))
+func ConfigMap(_ v1alpha1.AmazonCloudWatchAgent) string {
+	return "cwaagentconfig"
 }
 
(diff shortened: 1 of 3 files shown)


mitali-salvi (approved):

pkg/naming/main.go
```diff
@@ -9,8 +9,8 @@ import (
 )
 
 // ConfigMap builds the name for the config map used in the AmazonCloudWatchAgent containers.
-func ConfigMap(agent v1alpha1.AmazonCloudWatchAgent) string {
-	return DNSName(Truncate("%s", 63, agent.Name))
+func ConfigMap(_ v1alpha1.AmazonCloudWatchAgent) string {
+	return "cwaagentconfig"
```
lisguo, line 13: I guess this is okay, but this is also assuming that everyone will use the ACWA resource defined in helm. If we ever change that cm name then everything will break.

lisguo (approved):

pkg/naming/main.go
sky333999: Yeah this is just for the config map that the operator creates and mounts. Users currently shouldnt be even aware of this currently and cannot override it without changing the operator name itself. So we are defining this behavior for the first time and I guess we are saying we will always call this what our current public docs do.

PR #33, opened 2023-11-02
Updated the integration tests with resources name changes in addon

nathalapooja:
*Description of changes:*
Updated integration tests with eks addon resources name change and changed the default addon name and version to amazon-cloudwatch-observability

Testing on Beta cluster: https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/6736118187

### DIFF
--- a/.github/workflows/eks-add-on-integ-test.yml
+++ b/.github/workflows/eks-add-on-integ-test.yml
@@ -11,12 +11,12 @@ on:
       addon_name:
         required: true
         type: string
-        default: "amazon-cloudwatch"
+        default: "amazon-cloudwatch-observability"
         description: "EKS addon name"
       addon_version:
         required: true
         type: string
-        default: "v0.1.0-eksbuild.4"
+        default: "v1.1.0-eksbuild.1"
         description: "EKS addon version"
 
 concurrency:
--- a/integration-tests/eks-addon/validateResources_test.go
+++ b/integration-tests/eks-addon/validateResources_test.go
@@ -60,14 +60,8 @@ func TestK8s(t *testing.T) {
 	assert.Equal(t, v1.PodRunning, pods.Items[1].Status.Phase)
 	assert.Equal(t, v1.PodRunning, pods.Items[2].Status.Phase)
 
-	if validateAgentPodRegexMatch(pods.Items[0].Name) {
-		assert.True(t, validateOperatorPodRegexMatch(pods.Items[1].Name))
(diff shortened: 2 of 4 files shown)


integration-tests/eks-addon/validateResources_test.go
```diff
@@ -77,10 +71,10 @@ func TestK8s(t *testing.T) {
 		fmt.Println("service name: " + service.Name + " namespace:" + service.Namespace)
 	}
 	assert.Len(t, services.Items, 4)
-	assert.Equal(t, "amazon-cloudwatch-agent", services.Items[0].Name)
-	assert.Equal(t, "amazon-cloudwatch-agent-headless", services.Items[1].Name)
-	assert.Equal(t, "amazon-cloudwatch-agent-monitoring", services.Items[2].Name)
-	assert.Equal(t, "amazon-cloudwatch-agent-operator-webhook-service", services.Items[3].Name)
+	assert.Equal(t, "amazon-cloudwatch-observability-webhook-service", services.Items[0].Name)
+	assert.Equal(t, "cloudwatch-agent", services.Items[1].Name)
```
lisguo, line 75: nit: all of these values can be consts so if they change in the future, it will be easier to make modifications

lisguo (approved):

.github/workflows/eks-add-on-integ-test.yml
```diff
@@ -11,12 +11,12 @@ on:
       addon_name:
         required: true
         type: string
-        default: "amazon-cloudwatch"
+        default: "amazon-cloudwatch-observability"
         description: "EKS addon name"
       addon_version:
         required: true
         type: string
-        default: "v0.1.0-eksbuild.4"
```
mitali-salvi, line 19: will we need this update this for every add-on release

integration-tests/eks-addon/validateResources_test.go
mitali-salvi: +1

.github/workflows/eks-add-on-integ-test.yml
nathalapooja: This is to just provide default , when we execute this integration test, the MCM operator need to pass the releasing the addon version as workflow input to validate

mitali-salvi: gotcha thanks

mitali-salvi (approved):

PR #38, opened 2023-11-09
Add integration test for Helm charts

movence:
*Description of changes:*
Add integration tests that are similar to ESK add-on for Helm charts. This PR also consolidates add-on and helm charts tests into single test function since they validates the same attributes and components. 

Changes:
- add terraform integ test for Helm
- deletes EKS add-on test function then use consolidated EKS test function for both addon and helm

Test run: 
- https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/6850938986

### DIFF
--- a/.github/workflows/helm-integ-test.yml
+++ b/.github/workflows/helm-integ-test.yml
@@ -1,0 +1,98 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Run Helm Charts Integration Tests
+on:
+  workflow_dispatch:
+    inputs:
+      cluster_name:
+        required: true
+        type: string
+        default: "cwagent-operator-helm-integ"
+        description: "EKS cluster name"
+      cluster_namespace:
+        required: true
+        type: string
+        default: "amazon-cloudwatch"
+        description: "EKS cluster namespace"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+permissions:
+  id-token: write
+  contents: read
+
+env:
+  TERRAFORM_AWS_ASSUME_ROLE: ${{ secrets.TERRAFORM_AWS_ASSUME_ROLE }}
+  AWS_DEFAULT_REGION: us-west-2
+
+jobs:
+  HelmChartsIntegrationTest:
+    name: HelmChartsIntegrationTest
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      
+      - name: Generate testing id
+        run: echo TESTING_ID="${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v2
+        with:
+          role-to-assume: ${{ env.TERRAFORM_AWS_ASSUME_ROLE }}
+          aws-region: ${{ env.AWS_DEFAULT_REGION }}
+      
+      #  local directory to store the kubernetes config
+      - name: Create kubeconfig directory
+        run: mkdir -p ${{ github.workspace }}/.kube
+
+      - name: Set KUBECONFIG environment variable
+        run: echo KUBECONFIG="${{ github.workspace }}/.kube/config" >> $GITHUB_ENV
+
+      # - name: Set up kubeconfig
+      #   run: aws eks update-kubeconfig --name ${{ inputs.test-cluster-name }} --region ${{ env.AWS_DEFAULT_REGION }}
+
+      - name: Install eksctl
+        run: |
+          mkdir ${{ github.workspace }}/eksctl
+          curl -sLO "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_Linux_amd64.tar.gz"
+          tar -xzf eksctl_Linux_amd64.tar.gz -C ${{ github.workspace }}/eksctl && rm eksctl_Linux_amd64.tar.gz
+          echo "${{ github.workspace }}/eksctl" >> $GITHUB_PATH
+
+      - name: Verify Terraform version
+        run: terraform --version
+
+      - name: Terraform apply
+        uses: nick-fields/retry@v2
+        with:
+          max_attempts: 1
+          timeout_minutes: 60 # EKS takes about 20 minutes to spin up a cluster and service on the cluster
+          retry_wait_seconds: 5
+          command: |
+            cd integration-tests/terraform/helm
+            terraform init
+            if terraform apply -auto-approve \
+                -var="cluster_name=${{ github.event.inputs.cluster_name }}" \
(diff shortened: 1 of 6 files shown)


mitali-salvi (approved):

.github/workflows/helm-integ-test.yml
```diff
@@ -0,0 +1,95 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: Run Helm Charts Integration Tests
+on:
+  workflow_dispatch:
```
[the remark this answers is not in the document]
nathalapooja: This fails on push to main because the workflow dispatch event have to be triggered to get the workflow inputs.

.github/workflows/helm-integ-test.yml
```diff
@@ -0,0 +1,98 @@
...
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      cluster_name:
+        required: true
+        type: string
+        default: "cwagent-operator-helm-integ"
```
nathalapooja, line 14: Any specific reason to pass the cluster name instead of creating one as part integ test?

.github/workflows/helm-integ-test.yml
```diff
@@ -0,0 +1,98 @@
...
+        description: "EKS cluster name"
+      cluster_namespace:
+        required: true
+        type: string
+        default: "amazon-cloudwatch"
```
nathalapooja, line 19: Resources using helm will always be created in this namespace. Any specific reason to add as input instead of constant?

.github/workflows/helm-integ-test.yml
nathalapooja: This  default cluster might be deleted by someone in our test account and this integ test will fail in that case.

integration-tests/eks/validateResources_test.go
```diff
@@ -52,29 +57,34 @@ func TestK8s(t *testing.T) {
...
+	assert.Len(t, pods.Items, 3)
 	for _, pod := range pods.Items {
 		fmt.Println("pod name: " + pod.Name + " namespace:" + pod.Namespace)
+		assert.Equal(t, v1.PodRunning, pod.Status.Phase)
+		// matches
+		// - cloudwatch-agent-*
+		// - amazon-cloudwatch-observability-controller-manager-*
+		// - fluent-bit-*
+		if match, _ := regexp.MatchString(podNameRegex, pod.Name); !match {
+			assert.Fail(t, "Cluster Pods are not created correctly by Helm Charts")
```
nathalapooja, line 69: We are mentioning it as by Helm Charts but this validation will be used by EKS addon as well. This might be confusing.

integration-tests/eks/validateResources_test.go
```diff
@@ -92,61 +104,64 @@ func TestK8s(t *testing.T) {
 	//Validating the Daemon Sets
 	daemonSets, err := ListDaemonSets(nameSpace, clientSet)
 	assert.NoError(t, err)
+	assert.Len(t, daemonSets.Items, 2)
 	for _, daemonSet := range daemonSets.Items {
 		fmt.Println("daemonSet name: " + daemonSet.Name + " namespace:" + daemonSet.Namespace)
+		// matches
+		// - cloudwatch-agent
+		// - fluent-bit
+		if match, _ := regexp.MatchString(agentName+"|fluent-bit", daemonSet.Name); !match {
```
nathalapooja, line 113: Can you run the EKS addon integ test after these changes

.github/workflows/helm-integ-test.yml
```diff
@@ -0,0 +1,98 @@
...
+      - name: Create kubeconfig directory
+        run: mkdir -p ${{ github.workspace }}/../../../.kube
+
+      - name: Set KUBECONFIG environment variable
+        run: echo KUBECONFIG="${{ github.workspace }}/../../../.kube/config" >> $GITHUB_ENV
+
+      - name: Install eksctl
+        run: |
+          mkdir ${{ github.workspace }}/eksctl
+          curl -sLO "https://github.com/weaveworks/eksctl/releases/latest/download/eksctl_Linux_amd64.tar.gz"
```
nathalapooja, line 64: where are we using eksctl ?

.github/workflows/helm-integ-test.yml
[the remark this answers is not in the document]
movence: I'm not quite following. You are saying push to main triggers workflow dispatch? These values will be defaulted in the terraform, so I don't think they will fail.

.github/workflows/helm-integ-test.yml
movence: It's just a prefix to a cluster name which should help with easier debugging. It's just another way not hard coding it in the terraform resource block.

integration-tests/eks/validateResources_test.go
movence: good catch, will update it

.github/workflows/helm-integ-test.yml
movence: Are you sure helm resources use `amazon-cloudwatch` as the default namespace? I saw helm resources being created under default when the namespace is not given with `helm_release`. Maybe there is add-on configuration to deploy helm resources to `amazon-cloudwatch` namespace?

.github/workflows/helm-integ-test.yml
movence: removed

integration-tests/eks/validateResources_test.go
movence: https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/6895006518

I had to create a new branch to run not-yet-merged GHA, but it's just a new branch off of this PR with  a change to run on a push to that temp branch.

.github/workflows/helm-integ-test.yml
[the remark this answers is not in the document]
nathalapooja: Push to main doesn't trigger the workflow dispatch, so it will fail since the workflow inputs are not defaulted.

.github/workflows/helm-integ-test.yml
nathalapooja: For me its confusing as it says the cluster name and we are passing it to test. Can we just maintain similar to other integ tests where we just include as part of terraform?

.github/workflows/helm-integ-test.yml
nathalapooja: Makes sense.  Are we including this integ test as part of some MCM or preflight where we validate it for two namespaces default and amazon-cloudwatch? Can you help me in understanding the usecase of making it a workflow input?

.github/workflows/helm-integ-test.yml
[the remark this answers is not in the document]
nathalapooja: If push to main is already validated and if it considers the workflow inputs default, I am good.

.github/workflows/helm-integ-test.yml
movence: On second thought, we don't see any strong reason why we would want to use a different namespace. removed it

nathalapooja (approved):

mitali-salvi (approved):

PR #50, opened 2023-12-09
New build operator integration test when pushed to main

bhanuba:
**Description of the issue**
We currently have EKS addon integration tests, but nothing to test new builds of the operator that runs on push to main. So we are creating a test for the operator when pushed to main.

**Description of changes**

Creating Cloudwatch Agent daemonset with checking if it is running or not.
Creating a pod with java annotation and checking if default environmental variables are injected or not.
Creating a resource of kind instrumentation and checking environment variables are injected in pod within same namespace if jave instumentation is enable.
**License**

**Tests**
Integration Test: https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/7198738382/job/19608857980?pr=50


.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,84 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: OperatorIntegrationTest.yml
+on:
+  push:
+    branches:
+      - lisaguo-op-integ
```
lisguo, line 8: can we change this to execute on pull request?

.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,84 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: OperatorIntegrationTest.yml
```
lisguo, line 4: Can we change the name of this workflow to follow the same convention as the others?

`operator-integration-test.yml`

.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,84 @@
...
+          docker images
+
+      - name: Deploy operator to minikube
+        run:
+          make deploy
+
+      - name: Test case for AmagonCloudWatchAgent pod creation
+        run: |
+          cd integration-tests/operator
+          kubectl apply -f ./amazoncloudwatchagent.yaml
```
lisguo, line 56: nit: You can actually use helm to apply the template for amazoncloudwatchagent by using this one-liner:

```
helm template --namespace amazon-cloudwatch -s templates/cloudwatch-agent-daemonset.yaml ./helm | kubectl apply --namespace amazon-cloudwatch -f -
```

then we don't need the amazoncloudwatchagent.yaml at all

.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,84 @@
...
+          sleep 60
+          kubectl get pods -A
```
lisguo, line 58: probably don't need `kubectl get pods -A` here if we are gonna get pods of the namespace the next line

.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,84 @@
...
+          kubectl get pods -n amazon-cloudwatch
```
lisguo, line 59: nit: would it be better to have a go script to check if the cloudwatch agent is running? Similar to what we did for the namespace tests.

.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,84 @@
...
+            echo "Pod is created test case: PASS"
+          fi
+
+      - name: Test default instrumentation
+        run: |
+          cd integration-tests/operator
+          kubectl apply -f ./test-deployment.yaml
+          sleep 60
+          cd ./cmd
+          go run validate_instrumentation_vars.go default default_instrumentation_env_variables.json
```
lisguo, line 76: nit: would be cleaner to use absolute path here `go run cmd/validate_instrumentation_vars.go default default_instrumentation_env_variables.json`

.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,84 @@
...
+      - name: Test namespace instrumentation
+        run: |
+          cd integration-tests/operator
+          kubectl apply -f ./sample-instrumentation.yaml
+          kubectl apply -f ./ns-test-deployment.yaml
```
lisguo, line 81: We don't need this file. We can just have the `sample-instrumentation.yaml` be created in the `default` namespace.

Then we can restart the test deployment with
```
kubectl rollout restart deployment nginx
```

lisguo (changes requested):

.github/workflows/OperatorIntegrtionTest.yml
bhanuba: the status is not always running it is erroring out as we are running on minikube.

.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,90 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: operator-integration-test.yml
```
lisguo, line 4: the file yaml name has to change as well

.github/workflows/OperatorIntegrtionTest.yml
```diff
@@ -0,0 +1,90 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [ opened, reopened ]
+    branches:
+      - main
```
lisguo, line 12: nice!

integration-tests/operator/amazoncloudwatchagent.yaml
```diff
@@ -0,0 +1,71 @@
+apiVersion: cloudwatch.aws.amazon.com/v1alpha1
```
lisguo, line 1: we can delete this file now that we are using helm templates

integration-tests/operator/ns-test-deployment.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: apps/v1
```
lisguo, line 1: we can delete this file too

integration-tests/operator/cmd/validate_instrumentation_vars.go
```diff
@@ -0,0 +1,107 @@
...
+		os.Exit(1)
+	} else {
+		fmt.Println("Instrumentation Annotation Injection Test: PASS")
+	}
+}
+
+func verifyInstrumentationEnvVariables(namespace string, jsonpath string) bool {
+
+	var args []string
+	if namespace != "default" {
```
lisguo, line 29: this if/else statement can be removed since we are passing in the namespace

so we can just do
```
args = []string{"get", "pods", "-n", namespace, "-l", "app=nginx", "-o=jsonpath='{.items[*].metadata.name}'"}
```

integration-tests/operator/amazoncloudwatchagent.yaml
bhanuba: done!!

integration-tests/operator/ns-test-deployment.yaml
bhanuba: done!!

lisguo (dismissed):

.github/workflows/operator-integration-test.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Build image
+        run: |
+          eval $(minikube docker-env)
+          make container
+          docker images
+
+      - name: Deploy operator to minikube
+        run:
+          make deploy
```
nathalapooja, line 55: I see make deploy is running the below command: 
set-image-controller
	$(KUSTOMIZE) build config/default | kubectl apply -f -

does it deploy the public ecr image on the minicube ?

.github/workflows/operator-integration-test.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Verify minikube and cert-manager
+        run: |
+          sleep 10
+          kubectl get pods -A
+
+      - name: Build image
+        run: |
+          eval $(minikube docker-env)
+          make container
+          docker images
```
nathalapooja, line 51: I see in the build run , below command was executed to docker build
docker buildx build --load --platform linux/amd64 -t public.ecr.aws/cloudwatch-agent/cloudwatch-agent-operator:1.0.2 --build-arg VERSION_PKG="github.com/aws/amazon-cloudwatch-agent-operator/internal/version" --build-arg VERSION=1.0.2 --build-arg VERSION_DATE=2023-12-09T05:03:05Z --build-arg AGENT_VERSION="1.300031.1b317" --build-arg AUTO_INSTRUMENTATION_JAVA_VERSION="v1.31.1" .

Any reason to build this to public ecr image?

.github/workflows/operator-integration-test.yml
bhanuba: yes we are using minikube and public ecr image gets deployed to the minikube.

.github/workflows/operator-integration-test.yml
```diff
@@ -0,0 +1,90 @@
...
+        run: |
+          helm template --namespace amazon-cloudwatch -s templates/cloudwatch-agent-daemonset.yaml ./helm | kubectl apply --namespace amazon-cloudwatch -f -
+          sleep 60
+          kubectl describe pods -n amazon-cloudwatch
+          pod_name="$(kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/component=amazon-cloudwatch-agent,app.kubernetes.io/instance=amazon-cloudwatch.cloudwatch-agent -o=jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}')"
+          if [ -z "$pod_name" ]; then
+            echo "Pod $pod_name is not running. Exiting with ERROR."
+            echo "Pod is not created test case: FAIL"
+          else
+            echo "Pod $pod_name is running."
```
sky333999, line 67: Where are we checking that it is running?

.github/workflows/operator-integration-test.yml
```diff
@@ -0,0 +1,90 @@
...
+            echo "Pod is created test case: PASS"
+          fi
+
+      - name: Test default instrumentation
+        run: |
+          cd integration-tests/operator
```
sky333999, line 73: nit: I would call the directory `manifests` instead of `operator`. And call the file `sample-deployment.yaml` instead of `test-deployment.yaml` so its extensible for future tests.

.github/workflows/operator-integration-test.yml
lisguo: this is just the name of the container image after we run `make container`. When we deploy to minikube, it will take preference to the local image from `make container` which is the public ecr name. 

if this is too confusing, we can change `make container` to output something like `cloudwatch-agent-operator` and it would still work.

Makefile
```diff
@@ -1,5 +1,5 @@
 # Versions from versions.txt
-VERSION ?= "$(shell grep -v '\#' versions.txt | grep operator= | awk -F= '{print $$2}')"
+VERSION ?= $(shell grep -v '\#' versions.txt | grep operator= | awk -F= '{print $$2}')
```
sky333999, line 2: Was there a reason this had to change?

config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: 506463145083.dkr.ecr.us-west-2.amazonaws.com/cwagent-operator-pre-release
-  newTag: latest
+  newName: public.ecr.aws/cloudwatch-agent/cloudwatch-agent-operator
+  newTag: 1.0.2
```
sky333999, line 8: Is this being used for `make deploy` as well? 
Did we actually push to our public repo while running these github tests? Id imagine it atleast complains about permission issues. Mind checking?

.github/workflows/operator-integration-test.yml
bhanuba: Here we are not actually using the public image. We are building a new image with "make container" with public ecr name.

config/manager/kustomization.yaml
sky333999: We shouldnt be using our public repo and tags here.

.github/workflows/operator-integration-test.yml
```diff
@@ -0,0 +1,90 @@
...
+          kubectl apply -f ./test-deployment.yaml
+          sleep 60
+          kubectl get pods -A 
+          kubectl describe pods -n default
+          cd ./cmd
+          go run validate_instrumentation_vars.go default default_instrumentation_env_variables.json
```
sky333999, line 79: We are introducing a new pattern here that is diff from all the other integ tests we have currently that leverage tf. Would have been nice to just stay consistent.

integration-tests/operator/cmd/validate_instrumentation_vars.go
```diff
@@ -0,0 +1,103 @@
...
+		os.Exit(1)
+	} else {
+		fmt.Println("Instrumentation Annotation Injection Test: PASS")
+	}
+}
+
+func verifyInstrumentationEnvVariables(namespace string, jsonpath string) bool {
+
+	var args []string
+	args = []string{"get", "pods", "-n", namespace, "-l", "app=nginx", "-o=jsonpath='{.items[*].metadata.name}'"}
```
sky333999, line 29: Using the k8s go client would have been a lot cleaner. You would need to do all the unmarshalling etc. and now we have a weird mix of shell and go.

.github/workflows/operator-integration-test.yml
nathalapooja: I believe its better to use other name from public ecr image, it makes me assume that we are testing the public ecr image directly

.github/workflows/operator-integration-test.yml
bhanuba: We are running on the minikube. It will eventually fail. So we are checking with the name of the pod being created.

.github/workflows/operator-integration-test.yml
bhanuba: done!!

Makefile
bhanuba: Tested by giving the static value. Will change it to as it is.

.github/workflows/operator-integration-test.yml
sky333999: Why does the agent fail? What is the error message?
If the agent isnt running, we shouldnt print that the pod is running.

.github/workflows/operator-integration-test.yml
lisguo: I think this test is more lightweight so that we do not have to use terraform.

This is a test that can be used on every PR that spans about 6 minutes or so to quickly test the functionality of the operator.

We have the EKS / helm / E2E test that fully tests the whole system which uses terraform.

I like having the simplicity here imo.

Makefile
lisguo: This had to change because the latest tag would have quotes around it. So the image name ended up being `public.ecr.....:"latest"`

.github/workflows/operator-integration-test.yml
lisguo: that's fair, we would have to just change the name in the Makefile and kustomize.

lisguo (approved):

.github/workflows/operator-integration-test.yml
lisguo: It fails on minikube due to IMDS concerns. But the agent being healthy isn't really in scope for the operator integration tests IMO -- we just need to test the operator webhook functionality.

The caveat would be if we add a healthcheck in the operator...

.github/workflows/operator-integration-test.yml
sky333999: Doing it via terraform shouldnt change the runtime (or atleast by a significant amount). We can use the same minikube resource with terraform if we want to avoid bringing up an EKS cluster. But doing so already has comprises like being unable to verify if the agent is functional.

I think we are very shortly going to add health checks so itll be interesting how we handle that then.

.github/workflows/operator-integration-test.yml
sky333999: I think we plan on adding those health checks pretty soon.. so yeah this is bound to fail soon.

.github/workflows/operator-integration-test.yml
bhanuba: changed to aws/cloudwatch-agent-operator:dev

.github/workflows/operator-integration-test.yml
```diff
@@ -0,0 +1,92 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: operator Integration Test
+on:
+  push:
+    branches:
+      - main
+      - lisaguo-op-integ
```
nathalapooja, line 9: Lets remove this before merging to main

.github/workflows/operator-integration-test.yml
nathalapooja: From previous Lisa comments I see that we are not deploying the public ECR image.

integration-tests/operator/cmd/validate_instrumentation_vars.go
nathalapooja: I agree. We can make more similar to other integ test by using k8s client

Makefile
```diff
@@ -1,12 +1,12 @@
 # Versions from versions.txt
-VERSION ?= "$(shell grep -v '\#' versions.txt | grep operator= | awk -F= '{print $$2}')"
+VERSION ?= dev
```
nathalapooja, line 2: I assume this affects the actual image build flow for public

config/manager/kustomization.yaml
bhanuba: No we are not using this in make deploy. I have updated it now.

Makefile
lisguo: ahh, yeah I'd prefer we leave this to what is in versions.txt ... this defines the default images used for the operator if one is not provided.

integration-tests/manifests/sample-instrumentation.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: sample-instrumentation
+spec:
+  exporter:
+    endpoint: "http://example.com/collector"
+  sampler:
+    type: traceidratio
+    argument: "0.5"
+  resource:
+    addK8sUIDAttributes: true
+    resourceAttributes:
```
lisguo, line 13: not sure if you need this

.github/workflows/operator-integration-test.yml
lisguo: that's fair, we can use a linux ec2 for example.

this actually brings up a good point though -- is cwa only supported on AWS resources? I thought it would be able to run on native k8s?

.github/workflows/operator-integration-test.yml
```diff
@@ -0,0 +1,91 @@
...
+        run: |
+          eval $(minikube docker-env)
+          make container
+          docker images
+
+      - name: Deploy operator to minikube
+        run: | 
+          make deploy
+
+      - name: Test case for AmagonCloudWatchAgent pod creation
```
lisguo, line 57: typo: AmazonCloudWatchAgent

.github/workflows/operator-integration-test.yml
```diff
@@ -0,0 +1,91 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+name: operator Integration Test
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    types: [ opened, reopened ]
```
lisguo, line 10: can we include `synchronize, ready_for_review` ? 

This should trigger the build in this PR

lisguo (approved):

.github/workflows/operator-integration-test.yml
sky333999: It should be able to run on native k8s as well with the right creds config setup

sky333999 (approved):

sky333999:
I think there is a lot that can be cleaned up here to be consistent to our other tests, but its a non-blocking suggestion.

PR #51, opened 2023-12-15
E2E Test: Fix and update release testing

majanjua-amzn:
Change log:
- Fixing release testing by updating the override to work with EKS addon
- Updating test to use enablement script from aws-observability/application-signals-demo
- Added API calls before test run to speed up test like in other repos
- Removed `caller-workflow-name` input
- Pull test resources from aws-observability/aws-otel-java-instrumentation to remove code duplication
- Updating some of the versions for actions like checkout to better match the workflow in the ADOT repo
- Updated the image verification steps to show the operator image that is being used in the pods
- Updated the image verification steps to fail if the operator image did not change after patching

The change to patch the deployment is based on template defined here: [link](https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/helm/templates/operator-deployment.yaml#L27)

Test fails if image hasn't changed: [run](https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/7253958018)
Test passes if image is changed to a functional one: [run](https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/7253994359)

### DIFF
--- a/.github/workflows/appsignals-e2e-test.yml
+++ b/.github/workflows/appsignals-e2e-test.yml
@@ -5,6 +5,7 @@ name: App Signals Enablement E2E Testing
 on:
   workflow_call:
     inputs:
+      # Ensure two tests do not run on the same cluster at the same time through GitHub Action concurrency
       test-cluster-name:
         required: true
         type: string
@@ -141,7 +142,12 @@ jobs:
 
       - name: Verify pod CWAgent image
         run: |
-          kubectl get pods -n amazon-cloudwatch --output json | \
+          kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/name=cloudwatch-agent -o json | \
+          jq '.items[0].status.containerStatuses[0].imageID'
+
+      - name: Verify pod CWAgent Operator image
+        run: |
+          kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/name=amazon-cloudwatch-observability -o json | \
           jq '.items[0].status.containerStatuses[0].imageID'
 
       - name: Get the sample app endpoint
--- a/.github/workflows/appsignals-e2e-test.yml
+++ b/.github/workflows/appsignals-e2e-test.yml
@@ -1,5 +1,9 @@
 # This is a reusable workflow for running the E2E test for App Signals.
 # It is meant to be called from another workflow.
+# This E2E test is responsible for validating setting up a sample application on an EKS cluster and enabling
+# App Signals using the staging image of the CloudWatch Agent Operator. It validates the generated telemetry
+# including logs, metrics, and traces, then cleans up the cluster. The testing resources can be found in the
+# ADOT java instrumentation repo: https://github.com/aws-observability/aws-otel-java-instrumentation/tree/main/testing
 # Read more about reusable workflows: https://docs.github.com/en/actions/using-workflows/reusing-workflows#overview
 name: App Signals Enablement E2E Testing
 on:
--- a/.github/workflows/appsignals-e2e-test.yml
+++ b/.github/workflows/appsignals-e2e-test.yml
@@ -8,9 +8,6 @@ on:
       test-cluster-name:
         required: true
         type: string
-      caller-workflow-name:
-        required: true
-        type: string
 
 permissions:
   id-token: write
@@ -19,31 +16,41 @@ permissions:
 env:
   AWS_DEFAULT_REGION: us-east-1
   TEST_ACCOUNT: ${{ secrets.APP_SIGNALS_E2E_TEST_ACCOUNT }}
-  ENABLEMENT_SCRIPT_S3_BUCKET: ${{ secrets.APP_SIGNALS_E2E_ONBOARDING_ZIP_S3_URI }}
   SAMPLE_APP_NAMESPACE: sample-app-namespace
   SAMPLE_APP_FRONTEND_SERVICE_IMAGE: ${{ secrets.APP_SIGNALS_E2E_SAMPLE_APP_FRONTEND_SERVICE_IMAGE }}
   SAMPLE_APP_REMOTE_SERVICE_IMAGE: ${{ secrets.APP_SIGNALS_E2E_SAMPLE_APP_REMOTE_SERVICE_IMAGE }}
   METRIC_NAMESPACE: AppSignals
   LOG_GROUP: /aws/appsignals/eks
   ECR_OPERATOR_STAGING_IMAGE: ${{ secrets.ECR_OPERATOR_STAGING_IMAGE }}
-  ECR_OPERATOR_RELEASE_IMAGE: ${{ secrets.ECR_OPERATOR_RELEASE_IMAGE }}
 
 jobs:
   appsignals-e2e-test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      # This step avoids code duplication for terraform templates and the validator
+      # To simplify, we get the entire repo
+      - name: Get testing resources from ADOT
+        uses: actions/checkout@v4
         with:
-          # Checkout e2e-test branch and get only the required resources for testing
-          ref: e2e-test
+          repository: aws-observability/aws-otel-java-instrumentation
+          ref: main
+
+      - name: Download enablement script
+        uses: actions/checkout@v4
+        with:
+          repository: aws-observability/application-signals-demo
+          ref: main
+          path: enablement-script
           sparse-checkout: |
-            test
+            scripts/eks/appsignals/enable-app-signals.sh
+            scripts/eks/appsignals/clean-app-signals.sh
+          sparse-checkout-cone-mode: false
 
       - name: Generate testing id
-        run: echo TESTING_ID="${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+        run: echo TESTING_ID="${{ env.AWS_DEFAULT_REGION }}-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
 
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v2
+        uses: aws-actions/configure-aws-credentials@v4
         with:
           role-to-assume: ${{ secrets.APP_SIGNALS_E2E_TEST_ROLE_ARN }}
           aws-region: ${{ env.AWS_DEFAULT_REGION }}
@@ -78,17 +85,18 @@ jobs:
           --approve
 
       - name: Set up terraform
-        uses: hashicorp/setup-terraform@v2
+        uses: hashicorp/setup-terraform@v3
         with:
           terraform_wrapper: false
 
       - name: Deploy sample app via terraform
-        working-directory: test/terraform/eks
+        working-directory: testing/terraform/eks
         run: |
           terraform init
           terraform validate
           terraform apply -auto-approve \
             -var="test_id=${{ env.TESTING_ID }}" \
+            -var="aws_region=${{ env.AWS_DEFAULT_REGION }}" \
             -var="kube_directory_path=${{ github.workspace }}/.kube" \
             -var="eks_cluster_name=${{ inputs.test-cluster-name }}" \
             -var="eks_cluster_context_name=$(kubectl config current-context)" \
@@ -98,20 +106,20 @@ jobs:
             -var="sample_remote_app_image=${{ env.SAMPLE_APP_REMOTE_SERVICE_IMAGE }}"
 
       # Enable App Signals on the test cluster
-      - name: Pull and unzip enablement script from S3
-        run: aws s3 cp ${{ env.ENABLEMENT_SCRIPT_S3_BUCKET }} . && unzip -j onboarding.zip
-
-      - name: Set the CW Agent Operator image to the staging image in the manifest
-        if: inputs.caller-workflow-name == 'build-and-upload-staging'
-        run: "sed -i 's#${{ env.ECR_OPERATOR_RELEASE_IMAGE }}#${{ env.ECR_OPERATOR_STAGING_IMAGE }}#g' app-signals.yaml"
-
       - name: Enable App Signals
+        working-directory: enablement-script/scripts/eks/appsignals
         run: |
           ./enable-app-signals.sh \
           ${{ inputs.test-cluster-name }} \
           ${{ env.AWS_DEFAULT_REGION }} \
           ${{ env.SAMPLE_APP_NAMESPACE }}
-          
+
+      - name: Patch the CloudWatch Agent Operator image and restart CloudWatch pods
+        run: |
+          kubectl patch deploy -n amazon-cloudwatch amazon-cloudwatch-observability-controller-manager --type='json' -p '[{"op": "replace", "path": "/spec/template/spec/containers/0/image", "value": "${{ env.ECR_OPERATOR_STAGING_IMAGE }}"}]'
+          kubectl delete pods --all -n amazon-cloudwatch
+          kubectl wait --for=condition=Ready pod --all -n amazon-cloudwatch
+
       # Application pods need to be restarted for the
       # app signals instrumentation to take effect
       - name: Restart the app pods
@@ -139,9 +147,10 @@ jobs:
       - name: Get the sample app endpoint
         run: |
           echo "APP_ENDPOINT=$(terraform output sample_app_endpoint)" >> $GITHUB_ENV
-        working-directory: test/terraform/eks
+        working-directory: testing/terraform/eks
 
       - name: Wait for app endpoint to come online
+        id: endpoint-check
         run: |
           attempt_counter=0
           max_attempts=30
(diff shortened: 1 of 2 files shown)


harrryr (dismissed):

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -19,31 +16,41 @@ permissions:
...
   AWS_DEFAULT_REGION: us-east-1
   TEST_ACCOUNT: ${{ secrets.APP_SIGNALS_E2E_TEST_ACCOUNT }}
-  ENABLEMENT_SCRIPT_S3_BUCKET: ${{ secrets.APP_SIGNALS_E2E_ONBOARDING_ZIP_S3_URI }}
   SAMPLE_APP_NAMESPACE: sample-app-namespace
   SAMPLE_APP_FRONTEND_SERVICE_IMAGE: ${{ secrets.APP_SIGNALS_E2E_SAMPLE_APP_FRONTEND_SERVICE_IMAGE }}
   SAMPLE_APP_REMOTE_SERVICE_IMAGE: ${{ secrets.APP_SIGNALS_E2E_SAMPLE_APP_REMOTE_SERVICE_IMAGE }}
   METRIC_NAMESPACE: AppSignals
   LOG_GROUP: /aws/appsignals/eks
   ECR_OPERATOR_STAGING_IMAGE: ${{ secrets.ECR_OPERATOR_STAGING_IMAGE }}
-  ECR_OPERATOR_RELEASE_IMAGE: ${{ secrets.ECR_OPERATOR_RELEASE_IMAGE }}
```
thpierce, line 29: I'm confused why we no longer need the release image and only use the staging image.

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -19,31 +16,41 @@ permissions:
...
+            scripts/eks/appsignals/clean-app-signals.sh
+          sparse-checkout-cone-mode: false
 
       - name: Generate testing id
-        run: echo TESTING_ID="${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
+        run: echo TESTING_ID="${{ env.AWS_DEFAULT_REGION }}-${{ github.run_id }}-${{ github.run_number }}" >> $GITHUB_ENV
 
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@v2
+        uses: aws-actions/configure-aws-credentials@v4
```
thpierce, line 53: There are a number of version changes in this PR. Whats the motivation for them? E.g. `actions/checkout`, etc.

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -78,17 +85,18 @@ jobs:
...
 
       - name: Set up terraform
-        uses: hashicorp/setup-terraform@v2
+        uses: hashicorp/setup-terraform@v3
         with:
           terraform_wrapper: false
 
       - name: Deploy sample app via terraform
-        working-directory: test/terraform/eks
+        working-directory: testing/terraform/eks
```
thpierce, line 93: Why are we changing the directory?

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -98,20 +106,20 @@ jobs:
...
         run: |
           ./enable-app-signals.sh \
           ${{ inputs.test-cluster-name }} \
           ${{ env.AWS_DEFAULT_REGION }} \
           ${{ env.SAMPLE_APP_NAMESPACE }}
-          
+
+      - name: Patch the CloudWatch Agent Operator image and restart CloudWatch pods
+        run: |
+          kubectl patch deploy -n amazon-cloudwatch amazon-cloudwatch-observability-controller-manager --type='json' -p '[{"op": "replace", "path": "/spec/template/spec/containers/0/image", "value": "${{ env.ECR_OPERATOR_STAGING_IMAGE }}"}]'
```
thpierce, line 119: How can we show that this command is working as expected? Are we able to explicitly check the image in the pods after restart?

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -156,55 +165,63 @@ jobs:
...
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/aws-sdk-call/
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/remote-service?ip=${{ env.REMOTE_SERVICE_POD_IP }}/
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/client-call/
+
       # Validation for app signals telemetry data
       - name: Call endpoint and validate generated EMF logs
         id: log-validation
-        working-directory: test/validator
-        run: ./gradlew run --args='-c log-validation.yml
+        if: steps.endpoint-check.outcome == 'success' && !cancelled()
```
thpierce, line 180: Does this just allow tests to fail fast?

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -156,55 +165,63 @@ jobs:
...
           done
 
+      # This steps increases the speed of the validation by creating the telemetry data in advance
+      - name: Call all test APIs
+        continue-on-error: true
+        run: |
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/outgoing-http-call/
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/aws-sdk-call/
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/remote-service?ip=${{ env.REMOTE_SERVICE_POD_IP }}/
+          curl -S -s -o /dev/null http://${{ env.APP_ENDPOINT }}/client-call/
```
thpierce, line 175: Where were these being invoked before this change? I don't see any old steps doing this.

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -156,55 +165,63 @@ jobs:
...
+
       # Validation for app signals telemetry data
       - name: Call endpoint and validate generated EMF logs
         id: log-validation
-        working-directory: test/validator
-        run: ./gradlew run --args='-c log-validation.yml
+        if: steps.endpoint-check.outcome == 'success' && !cancelled()
+        run: ./gradlew testing:validator:run --args='-c eks/log-validation.yml
```
thpierce, line 181: So was this just failing before because of a) incorrect command specification and b) incorrect path? Can you provide a failed test run for comparison?

Also is the command missing a `'`? e.g. `--args='-c eks/log-validation.yml` -> `--args='-c eks/log-validation.yml'`.

.github/workflows/build-and-upload-staging.yml
```diff
@@ -66,8 +66,7 @@ jobs:
     uses: ./.github/workflows/appsignals-e2e-test.yml
     secrets: inherit
     concurrency:
-      group: 'appsignals-cw-agent-operator-test'
+      group: 'e2e-cw-agent-operator-test'
```
thpierce, line 69: Why the rename?

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: We previously used the release image in our `sed` command to replace it with the staging image. We no longer need to do that since I used `kubectl patch` to change the value.

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: The same changes occurred in the ADOT repo, just matching them here. Forgot to add that in the PR description, will add it now.

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: We're pulling the testing resources from ADOT directly now, where the folder is named `testing`. We didn't get to choose the names because each repo already had a preexisting `test` or `testing` folder.

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: The current set up allows metric and trace validation to run whether or not the log validation failed. This is so that we can identify which part of the enablement experience is being impacted if failures occur.

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: They are being invoked in the validator (and still are), but this just helps speed it up because while gradle is building the package the telemetry has time to appear in CloudWatch.

.github/workflows/build-and-upload-staging.yml
majanjua-amzn: If we change the group name to the cluster name it's more clear that the limiting factor for concurrency is the cluster. I'll add a comment to ensure people use it properly

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: No, the test was previously silently failing because our solution to replace the CW Agent operator was deprecated due to use of the EKS addon instead of deployment configurations/instrumentation files. The reason it was silently failing was because `sed` commands don't have failure outputs unless the file it is targeting doesn't exist. As such, the release image was being tested which passes through the validation.

`'` is not missing, it's at the end of the `run:` value a few lines down. This string contains all the arguments needed to run the validator.

majanjua-amzn: See https://github.com/aws-observability/aws-otel-java-instrumentation/pull/652 for test runs that show the current logic will affect change in the outcome of the test. It is different because we implemented a complex retry mechanism on the enablement process for the canary.

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: I updated the image verification steps. Good callout

.github/workflows/appsignals-e2e-test.yml
thpierce: Got it.

.github/workflows/appsignals-e2e-test.yml
thpierce: Ack, reasonable.

.github/workflows/appsignals-e2e-test.yml
thpierce: Makes sense.

.github/workflows/appsignals-e2e-test.yml
thpierce: Spoke offline, it's for fail fast.

.github/workflows/appsignals-e2e-test.yml
thpierce: Makes sense.

.github/workflows/appsignals-e2e-test.yml
thpierce: `testing:validator:` and `eks/` added due to change in how test are setup and run. Makes sense.

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -133,15 +142,21 @@ jobs:
...
       - name: Verify pod CWAgent image
         run: |
-          kubectl get pods -n amazon-cloudwatch --output json | \
+          kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/name=cloudwatch-agent -o json | \
+          jq '.items[0].status.containerStatuses[0].imageID'
+
+      - name: Verify pod CWAgent Operator image
+        run: |
+          kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/name=amazon-cloudwatch-observability -o json | \
           jq '.items[0].status.containerStatuses[0].imageID'
```
thpierce, line 151: Lets actually validate the image id.

thpierce (commented):
Please do a new test run with the next revision to show the latest changes work as expected.

thpierce (approved):

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -19,31 +17,41 @@ permissions:
...
   appsignals-e2e-test:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      # This step avoids code duplication for terraform templates and the validator
+      # To simplify, we get the entire repo
+      - name: Get testing resources from ADOT
+        uses: actions/checkout@v4
+        with:
+          repository: aws-observability/aws-otel-java-instrumentation
```
lisguo, line 36: does the adot repo have a staging image to use? Curious if we might run into a race condition where we are building main of agent + adot where both have changes which break the build.

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -19,31 +17,41 @@ permissions:
...
+        uses: actions/checkout@v4
         with:
-          # Checkout e2e-test branch and get only the required resources for testing
-          ref: e2e-test
+          repository: aws-observability/application-signals-demo
+          ref: main
+          path: enablement-script
           sparse-checkout: |
-            test
+            scripts/eks/appsignals/enable-app-signals.sh
```
lisguo, line 46: should we have documentation somewhere or change the name to be `E2E test with demo scripts` of some sort? Can be confusing with all of our E2E tests with the EKS addon / helm / etc...

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -98,20 +107,25 @@ jobs:
...
         run: |
           ./enable-app-signals.sh \
           ${{ inputs.test-cluster-name }} \
           ${{ env.AWS_DEFAULT_REGION }} \
           ${{ env.SAMPLE_APP_NAMESPACE }}
-          
+
+      - name: Save CloudWatch Agent Operator image to environment before patching
+        run: |
+          echo "OLD_CW_AGENT_OPERATOR_IMAGE"=$(kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/name=amazon-cloudwatch-observability -o json | \
```
lisguo, line 120: this is getting the old operator image from the demo script right? Are we hardcoding the version in those scripts?

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -98,20 +107,25 @@ jobs:
...
+          jq '.items[0].status.containerStatuses[0].image') >> $GITHUB_ENV
+
+      - name: Patch the CloudWatch Agent Operator image and restart CloudWatch pods
+        run: |
+          kubectl patch deploy -n amazon-cloudwatch amazon-cloudwatch-observability-controller-manager --type='json' -p '[{"op": "replace", "path": "/spec/template/spec/containers/0/image", "value": "${{ env.ECR_OPERATOR_STAGING_IMAGE }}"}]'
+          kubectl delete pods --all -n amazon-cloudwatch
+          kubectl wait --for=condition=Ready pod --all -n amazon-cloudwatch
```
lisguo, line 127: nit: can probably make `amazon-cloudwatch` a const since we are using it everywhere

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -126,22 +140,38 @@ jobs:
...
           jq '.items[0].status.initContainerStatuses[0].imageID'
 
-      - name: Verify pod CWAgent image
+      - name: Log pod CWAgent image ID
         run: |
-          kubectl get pods -n amazon-cloudwatch --output json | \
+          kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/name=cloudwatch-agent -o json | \
           jq '.items[0].status.containerStatuses[0].imageID'
 
+      - name: Log pod CWAgent Operator image ID and save image to the environment
```
lisguo, line 153: The addon also installs fluentbit ... do we want to include that here as well? We are basically testing it already since it gets deployed to the `amazon-cloudwatch` namespace and we check that all pods are in the Ready state

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: They do, but that's not being tested here. We have release testing that uses a separate cluster and workflow in their repo. In this test we use the live version of the enablement script that has no staging changes to it and change only the operator.

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: Technically yes, but what it's really doing is getting the operator image in use after enabling pulse and before overriding it for the release testing. This is to ensure after we make our patch to the deployment that the operator image has changed.

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: If fluentbit has a specific version for pulse, then we can add a validation for it.

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: I believe we intend to vend the enablement script through that demo repo and not through any other means now. Also seems pretty self-explanatory what the use of this step is since we checkout the `enable-app-signals.sh` and `clean-app-signals.sh` scripts.

Maybe I misunderstood what needs to be clarified here through docs/comments?

lisguo: I was concerned with distinction between `appsignals-e2e-test` and our other integ tests. But I guess this is the only one called "E2E" so it's probably fine.

lisguo (dismissed):

nathalapooja (commented):
Can we add more description to the github workflow what exactly this test is doing?

.github/workflows/appsignals-e2e-test.yml
nathalapooja: Can we add for fluent-bit image as well

.github/workflows/appsignals-e2e-test.yml
```diff
@@ -126,22 +140,38 @@ jobs:
...
+        run: |
+          kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/name=amazon-cloudwatch-observability -o json | \
+          jq '.items[0].status.containerStatuses[0].imageID'
+          
+          echo "NEW_CW_AGENT_OPERATOR_IMAGE"=$(kubectl get pods -n amazon-cloudwatch -l app.kubernetes.io/name=amazon-cloudwatch-observability -o json | \
+          jq '.items[0].status.containerStatuses[0].image') >> $GITHUB_ENV
+
+      - name: Check if CW Agent Operator image has changed
+        run: |
+          if [ ${{ env.OLD_CW_AGENT_OPERATOR_IMAGE }} = ${{ env.NEW_CW_AGENT_OPERATOR_IMAGE }} ]; then
```
nathalapooja, line 167: We can add other condition ${{ env.NEW_CW_AGENT_OPERATOR_IMAGE }} != env.ECR_OPERATOR_STAGING_IMAGE

majanjua-amzn: I tried this and the statement would never pass for some reason so I went with this which accomplishes the same goal.

nathalapooja (approved):

.github/workflows/appsignals-e2e-test.yml
majanjua-amzn: Taking action to do this in a small separate PR

lisguo (approved):

PR #58, opened 2024-01-12
Added logic to parse the receiver ports from cw agent config json

nathalapooja:
*Issue #, if available:*

*Description of changes:*
Added logic to parse the receiver ports from cw agent config json.

Testing 
Unit Tests
Manual Testing with below Advanced EKS addon Cw agent config:
```
{
    "agent": {
        "config": {
            "agent": {
                "debug": true
            },
            "metrics": {
                "metrics_collected": {
                    "statsd": {
                        "service_address": ":8135"
                    }
                }
            },
            "logs": {
                "metrics_collected": {
                    "kubernetes": {
                        "enhanced_container_insights": true
                    },
                    "app_signals": {},
                    "emf": {}
                }
            },
            "traces": {
                "traces_collected": {
                    "app_signals": {},
                    "xray": {
                        "bind_address": "127.0.0.1:2800",
                        "tcp_proxy": {
                            "bind_address": "127.0.0.1:2900"
                        }
                    },
                    "otlp": {
                        "grpc_endpoint": "127.0.0.1:4327",
                        "http_endpoint": "127.0.0.1:4328"
                    }
                }
            }
        }
    }
}
```

Pod Spec of Cloudwatch agent pod:

```
poojardy@88665a4a5be5 ~ % kubectl describe pod cloudwatch-agent-7z7tp -n amazon-cloudwatch
Containers:
  cloudwatch-agent:
    Container ID:   containerd://9f96aaaee3dbf5d6dcb569a1f5ac415345cb7e3425b91df2e4ca5e701bccd1f8
    Image:          public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.300031.1b317
    Image ID:       public.ecr.aws/cloudwatch-agent/cloudwatch-agent@sha256:d660a92f0e97e529235e21d0ee440942f503c4cdd48dadaa80781fdc11427087
    Ports:          4315/TCP, 4316/TCP, 2000/TCP, 2900/TCP, 2800/UDP, 4327/TCP, 4328/TCP, 8135/UDP, 25888/TCP
    Host Ports:     0/TCP, 0/TCP, 0/TCP, 0/TCP, 0/UDP, 0/TCP, 0/TCP, 0/UDP, 0/TCP
    State:          Running
      Started:      Fri, 12 Jan 2024 15:23:27 -0500
    Ready:          True
    Restart Count:  0
    Limits:
      cpu:     500m
      memory:  512Mi
    Requests:
      cpu:     250m
      memory:  128Mi
```

### DIFF
--- a/pkg/collector/container.go
+++ b/pkg/collector/container.go
@@ -106,20 +106,18 @@ func Container(cfg config.Config, logger logr.Logger, agent v1alpha1.AmazonCloud
 		Env:             envVars,
 		EnvFrom:         agent.Spec.EnvFrom,
 		Resources:       agent.Spec.Resources,
-		Ports:           portMapToList(ports),
+		Ports:           portMapToContainerPortList(ports),
 	}
 }
 
 func getContainerPorts(logger logr.Logger, cfg string) map[string]corev1.ContainerPort {
 	ports := map[string]corev1.ContainerPort{}
 	var servicePorts []corev1.ServicePort
-	servicePorts = append(servicePorts, AppSignalsCloudwatchAgentPorts...)
-
 	configJSON, err := adapters.ConfigStructFromJSONString(cfg)
 	if err != nil {
 		logger.Error(err, "error parsing cw agent config")
 	} else {
-		servicePorts = getServicePorts(configJSON, servicePorts)
+		servicePorts = getServicePortsFromCWAgentConfig(configJSON)
 	}
 
 	for _, p := range servicePorts {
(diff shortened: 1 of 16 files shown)


pkg/collector/container.go
```diff
@@ -129,7 +143,108 @@ func getContainerPorts(logger logr.Logger, cfg string) map[string]corev1.Contain
...
+	if configJSON.Metrics.MetricsCollected.CollectD != nil {
+		getReceiverServicePort(logger, configJSON.Metrics.MetricsCollected.CollectD.ServiceAddress, CollectD, corev1.ProtocolUDP, containerPortsMap)
+	}
+}
+
+func getReceiverServicePort(logger logr.Logger, serviceAddress string, receiverName string, protocol corev1.Protocol, containerPortsMap map[int32]corev1.ServicePort) {
+	if serviceAddress != "" {
+		port, err := portFromEndpoint(serviceAddress)
+		if err != nil {
+			logger.Error(err, "error parsing port from endpoint")
```
lisguo, line 172: nit: would be good to log the receiver name and endpoint

pkg/collector/container.go
```diff
@@ -129,7 +143,108 @@ func getContainerPorts(logger logr.Logger, cfg string) map[string]corev1.Contain
...
+	if configJSON.Metrics.MetricsCollected.StatsD != nil {
+		getReceiverServicePort(logger, configJSON.Metrics.MetricsCollected.StatsD.ServiceAddress, StatsD, corev1.ProtocolUDP, containerPortsMap)
+	}
+	//CollectD - https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-custom-metrics-collectd.html
+	if configJSON.Metrics.MetricsCollected.CollectD != nil {
+		getReceiverServicePort(logger, configJSON.Metrics.MetricsCollected.CollectD.ServiceAddress, CollectD, corev1.ProtocolUDP, containerPortsMap)
+	}
+}
+
+func getReceiverServicePort(logger logr.Logger, serviceAddress string, receiverName string, protocol corev1.Protocol, containerPortsMap map[int32]corev1.ServicePort) {
```
lisguo, line 168: it's not a good practice to modify the input parameters. Can we just return the ServicePort object and then the caller will modify the containerPortsMap?

pkg/collector/container.go
lisguo, line 172: if there is an error parsing the endpoint, should we fall back to using the default?

lisguo:
Also something worth noting -- the upstream opentelemetry-operator seems to use a config parser where each receiver is it's own component and has it's own scraping logic: https://github.com/open-telemetry/opentelemetry-operator/blob/main/internal/manifests/collector/parser/receiver/receiver.go

Maybe there is something we can borrow here? Not a blocker by any means since this work has already been completed.

pkg/collector/container.go
```diff
@@ -101,23 +106,32 @@ func Container(cfg config.Config, logger logr.Logger, agent v1alpha1.AmazonCloud
...
+		logger.Error(err, "error parsing cw agent config")
+		servicePorts = PortMapToServicePortList(AppSignalsPortToServicePortMap)
+	} else {
+		servicePorts = getServicePortsFromCWAgentConfig(logger, configJSON)
+	}
+
+	for _, p := range servicePorts {
 		truncName := naming.Truncate(p.Name, maxPortLen)
 		if p.Name != truncName {
 			logger.Info("truncating container port name",
```
mitali-salvi, line 127: np: can this be made in to debug instead of info ?

pkg/collector/container.go
mitali-salvi: +1

pkg/collector/container.go
mitali-salvi: I think yes ! I was also thinking what should happen if the customer supplies an invalid config (ex - statsd config without the service address field), should the operator open default ports ? We should mimic the CWA behavior in that case

nathalapooja: If they don't mention the service address at all in the config, code already handles to add the default address. But if the service address given in the config is not valid, I think it doesn't make sense to make it default as customer might not be using that port at all.

pkg/collector/container.go
nathalapooja: This logger of type logr.logger doesn't have debug function. Only Error and Info are supported

pkg/collector/container.go
nathalapooja: In Go, it is idiomatic to modify structures passed in as parameters rather than returning newly created ones, especially for larger data structures. This is because Go uses pointers to pass data structures, and modifying the structure in place can be more efficient than creating a new one and returning it.

mitali-salvi (approved):

lisguo (approved):

PR #64, opened 2024-01-24
Run CW agent as host process container

KlwntSingh:
*Description of changes:*
1. fix integration tests
2. Run CW agent as host process container

*Testing*
Tested integ run in github workflow action and successfuly passed

### DIFF
--- a/integration-tests/terraform/helm-windows/main.tf
+++ b/integration-tests/terraform/helm-windows/main.tf
@@ -196,9 +196,16 @@ resource "null_resource" "kubectl" {
   }
 }
 
+resource "time_sleep" "wait_15_min" {
+  depends_on = [helm_release.this]
+
+  create_duration = "15m"
+}
+
 resource "helm_release" "this" {
   depends_on = [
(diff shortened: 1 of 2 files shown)


integration-tests/terraform/helm-windows/main.tf
```diff
@@ -196,9 +196,16 @@ resource "null_resource" "kubectl" {
   }
 }
 
+resource "time_sleep" "wait_15_min" {
+  depends_on = [helm_release.this]
+
+  create_duration = "15m"
```
chadpatel, line 202: why 15 minutes? is that the same as we do for linux?

chadpatel: What is typical for linux? My concern here is we are going to end up with really long test executions.
How do the execution times compare for lin and win?
Is there another method? Polling?  Presumably the windows image only changes every ~month or so caching should help?

KlwntSingh:
Clsoing this PR since part of it is already merged.
Integ tests will be added as part of different PR.

PR #68, opened 2024-01-29
Add auto-annotation mutators on start up

jefchien:
*Issue #, if available:* N/A

*Description of changes:*
- Adds auto-annotation configuration that gets passed in through `auto-annotation-config` flag.
- Adds auto-annotation mutators that get run at start up.
  - `AnnotationMutator` is partially based on `kubectl annotate` implementation (https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/annotate/annotate.go#L457).
  - Has functions for each of the supported resources.
- Adds unit tests for annotation mutators.

### DIFF
--- a/main.go
+++ b/main.go
@@ -100,7 +100,7 @@ func main() {
 	pflag.StringVar(&pprofAddr, "pprof-addr", "", "The address to expose the pprof server. Default is empty string which disables the pprof server.")
 	stringFlagOrEnv(&agentImage, "agent-image", "RELATED_IMAGE_COLLECTOR", fmt.Sprintf("%s:%s", cloudwatchAgentImageRepository, v.AmazonCloudWatchAgent), "The default CloudWatch Agent image. This image is used when no image is specified in the CustomResource.")
 	stringFlagOrEnv(&autoInstrumentationJava, "auto-instrumentation-java-image", "RELATED_IMAGE_AUTO_INSTRUMENTATION_JAVA", fmt.Sprintf("%s:%s", autoInstrumentationJavaImageRepository, v.AutoInstrumentationJava), "The default OpenTelemetry Java instrumentation image. This image is used when no image is specified in the CustomResource.")
-	pflag.StringVar(&autoAnnotationConfigStr, "auto-annotation-config", "", "The configuration for auto-annotation.")
+	stringFlagOrEnv(&autoAnnotationConfigStr, "auto-annotation-config", "AUTO_ANNOTATION_CONFIG", "", "The configuration for auto-annotation.")
 	pflag.Parse()
 
 	// set java instrumentation java image in environment variable to be used for default instrumentation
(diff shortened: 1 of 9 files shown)


pkg/instrumentation/annotationmutator.go
```diff
@@ -0,0 +1,106 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package instrumentation
```
mitali-salvi, line 4: wondering it files related to annotation should be in the annotation package instead of instrumentation ??

jefchien: This is all in service of instrumentation. Similar to how the `podmutator` is in the `instrumentation` package.

pkg/instrumentation/annotationmutator.go
```diff
@@ -0,0 +1,106 @@
...
+	requireAll bool
+}
+
+func (m *insertAnnotationMutation) Mutate(annotations map[string]string) bool {
+	if m.requireAll && !m.validate(annotations) {
+		return false
+	}
+	var mutated bool
+	for key, value := range m.insert {
+		if _, ok := annotations[key]; !ok {
```
lisguo, line 27: do we need a mutex for read/write of the annotations map?

jefchien: Only if we're modifying the annotations on multiple threads. We aren't currently doing that, so I don't think we need it yet.

pkg/instrumentation/auto/annotation.go
```diff
@@ -0,0 +1,245 @@
...
+func (m *AnnotationMutators) MutateDeployment(ctx context.Context, deployment appsv1.Deployment) {
+	name := namespacedName(deployment.GetObjectMeta())
+	mutator, ok := m.deploymentMutators[name]
+	if !ok {
+		mutator = m.defaultMutator
+	}
+	if !mutator.Mutate(deployment.Spec.Template.GetObjectMeta()) {
+		return
+	}
+	if err := m.client.Update(ctx, &deployment); err != nil {
```
jefchien, line 94: If we're expecting to reuse this for the webhooks, the updates likely should be done outside of this function and the function should likely return the deployment and whether it was mutated.

jefchien: Though it looks like the webhooks don't care if the mutators made any changes, so maybe it doesn't need to return if it mutated the resource.

main.go
```diff
@@ -207,6 +211,23 @@ func main() {
 		os.Exit(1)
 	}
 
+	var autoAnnotationConfig auto.AnnotationConfig
+	if err := json.Unmarshal([]byte(autoAnnotationConfigStr), &autoAnnotationConfig); err != nil {
+		setupLog.Error(err, "unable to unmarshal auto-annotation config")
```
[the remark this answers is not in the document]
jefchien: We can wrap this in a recover if you're concerned about panics.

main.go
```diff
@@ -207,6 +211,23 @@ func main() {
 		os.Exit(1)
 	}
 
+	var autoAnnotationConfig auto.AnnotationConfig
+	if err := json.Unmarshal([]byte(autoAnnotationConfigStr), &autoAnnotationConfig); err != nil {
```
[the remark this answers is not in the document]
jefchien: If everything was unselected, I wouldn't expect to get an empty string, but it should still work either way. Added the `DISABLE_AUTO_ANNOTATION`.

main.go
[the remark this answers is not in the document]
jefchien: Will do this as part of a different PR once the webhooks have been added.

lisguo (approved):

sky333999 (approved):

mitali-salvi (approved):

PR #73, opened 2024-01-31
Implementing workload webhook for auto-annotation

mitali-salvi:
*Description of changes:*
Implements a workloads webhook (deployments, daemon-sets and statefulsets)
Add the update RBAC role for updating namespaces
add a clientReader to the annotation-mutator

*Testing*
For the config - {"java":{"namespaces":["test"],"daemonsets":["default/prometheus-daemonset"]}}

```
kubectl apply -f ds.yaml 
daemonset.apps/prometheus-daemonset created


kubectl describe ds
Name:           prometheus-daemonset
Selector:       name=prometheus-exporter,tier=monitoring
Node-Selector:  <none>
Labels:         <none>
Annotations:    deprecated.daemonset.template.generation: 1
Desired Number of Nodes Scheduled: 2
Current Number of Nodes Scheduled: 2
Number of Nodes Scheduled with Up-to-date Pods: 2
Number of Nodes Scheduled with Available Pods: 2
Number of Nodes Misscheduled: 0
Pods Status:  2 Running / 0 Waiting / 0 Succeeded / 0 Failed
Pod Template:
  Labels:       name=prometheus-exporter
                tier=monitoring
  Annotations:  cloudwatch.aws.amazon.com/auto-annotate-java: true
                instrumentation.opentelemetry.io/inject-java: true
  Containers:
   prometheus:
    Image:        prom/node-exporter
    Port:         80/TCP
    Host Port:    0/TCP
    Environment:  <none>
    Mounts:       <none>
  Volumes:        <none>
Events:
  Type    Reason            Age   From                  Message
  ----    ------            ----  ----                  -------
  Normal  SuccessfulCreate  7s    daemonset-controller  Created pod: prometheus-daemonset-64rt2
  Normal  SuccessfulCreate  7s    daemonset-controller  Created pod: prometheus-daemonset-24x48
```

Testing auto-annotate resources
```
kubectl create ns test                  
namespace/test created

kubectl rollout restart deploy cloudwatch-controller-manager -n amazon-cloudwatch
deployment.apps/cloudwatch-controller-manager restarted

kubectl describe ns test                                                    
Name:         test
Labels:       kubernetes.io/metadata.name=test
Annotations:  cloudwatch.aws.amazon.com/auto-annotate-java: true
              instrumentation.opentelemetry.io/inject-java: true
Status:       Active
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -32,6 +32,7 @@ rules:
   - namespaces
   verbs:
   - list
+  - update
   - watch
 - apiGroups:
   - apps
--- a/internal/webhook/workloadmutation/webhookhandler.go
+++ b/internal/webhook/workloadmutation/webhookhandler.go
@@ -51,15 +51,6 @@ func NewWebhookHandler(cfg config.Config, logger logr.Logger, decoder *admission
 }
 
 func (p *workloadMutationWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
-	if p.annotationMutator == nil {
-		// By default, admission.Errored sets Allowed to false which blocks workload creation even though the failurePolicy=ignore.
-		// Allowed set to true makes sure failure does not block workload creation in case of an error.
-		// Returning http.StatusBadRequest does not create any event.
-		res := admission.Errored(http.StatusBadRequest, errors.New("failed to unmarshal annotation config"))
-		res.Allowed = true
-		return res
-	}
-
 	var err error
 	var marshaledObject []byte
 	var object runtime.Object
--- a/internal/webhook/workloadmutation/webhookhandler.go
+++ b/internal/webhook/workloadmutation/webhookhandler.go
@@ -32,27 +32,32 @@ type WebhookHandler interface {
 
 // the implementation.
 type workloadMutationWebhook struct {
-	client             client.Client
-	decoder            *admission.Decoder
-	logger             logr.Logger
-	config             config.Config
-	annotationMutation *auto.AnnotationMutators
+	client            client.Client
+	decoder           *admission.Decoder
+	logger            logr.Logger
+	config            config.Config
+	annotationMutator *auto.AnnotationMutators
 }
 
 // NewWebhookHandler creates a new WorkloadWebhookHandler.
 func NewWebhookHandler(cfg config.Config, logger logr.Logger, decoder *admission.Decoder, cl client.Client, annotationMutation *auto.AnnotationMutators) WebhookHandler {
 	return &workloadMutationWebhook{
-		config:             cfg,
-		decoder:            decoder,
-		logger:             logger,
-		client:             cl,
-		annotationMutation: annotationMutation,
+		config:            cfg,
+		decoder:           decoder,
+		logger:            logger,
+		client:            cl,
+		annotationMutator: annotationMutation,
 	}
 }
 
 func (p *workloadMutationWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
-	if p.annotationMutation == nil {
-		return admission.Errored(http.StatusBadRequest, errors.New("failed to unmarshal annotation config"))
+	if p.annotationMutator == nil {
+		// By default, admission.Errored sets Allowed to false which blocks workload creation even though the failurePolicy=ignore.
+		// Allowed set to true makes sure failure does not block workload creation in case of an error.
+		// Returning http.StatusBadRequest does not create any event.
+		res := admission.Errored(http.StatusBadRequest, errors.New("failed to unmarshal annotation config"))
+		res.Allowed = true
+		return res
 	}
 
 	var err error
@@ -65,7 +70,6 @@ func (p *workloadMutationWebhook) Handle(ctx context.Context, req admission.Requ
 		if err != nil {
 			return admission.Errored(http.StatusBadRequest, err)
 		}
-		p.annotationMutation.MutateDaemonSet(&ds)
 		object = &ds
 	case "Deployment":
 		d := appsv1.Deployment{}
@@ -73,7 +77,6 @@ func (p *workloadMutationWebhook) Handle(ctx context.Context, req admission.Requ
 		if err != nil {
 			return admission.Errored(http.StatusBadRequest, err)
 		}
-		p.annotationMutation.MutateDeployment(&d)
 		object = &d
 	case "StatefulSet":
 		ss := appsv1.StatefulSet{}
@@ -81,12 +84,12 @@ func (p *workloadMutationWebhook) Handle(ctx context.Context, req admission.Requ
 		if err != nil {
 			return admission.Errored(http.StatusBadRequest, err)
 		}
-		p.annotationMutation.MutateStatefulSet(&ss)
 		object = &ss
 	default:
 		return admission.Errored(http.StatusBadRequest, errors.New("failed to unmarshal request object"))
 	}
 
+	p.annotationMutator.Mutate(object)
 	marshaledObject, err = json.Marshal(object)
 	if err != nil {
 		res := admission.Errored(http.StatusInternalServerError, err)
--- a/internal/webhook/workloadmutation/webhookhandler.go
+++ b/internal/webhook/workloadmutation/webhookhandler.go
@@ -1,0 +1,98 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package workloadmutation contains the webhook that injects annotations into daemon-sets, deployments and stateful-sets.
+package workloadmutation
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"net/http"
+
+	"github.com/go-logr/logr"
+	appsv1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/aws/amazon-cloudwatch-agent-operator/internal/config"
+	"github.com/aws/amazon-cloudwatch-agent-operator/pkg/instrumentation/auto"
+)
+
+// +kubebuilder:webhook:path=/mutate-v1-workload,mutating=true,failurePolicy=ignore,groups="apps",resources=daemonsets;deployments;statefulsets,verbs=create;update,versions=v1,name=mworkload.kb.io,sideEffects=none,admissionReviewVersions=v1
+// +kubebuilder:rbac:groups="apps",resources=daemonsets;deployments;statefulsets,verbs=get;list;watch
+
+var _ WebhookHandler = (*workloadMutationWebhook)(nil)
+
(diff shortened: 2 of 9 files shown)


internal/webhook/workloadmutation/webhookhandler.go
```diff
@@ -0,0 +1,98 @@
...
+	admission.Handler
+}
+
+// the implementation.
+type workloadMutationWebhook struct {
+	client             client.Client
+	decoder            *admission.Decoder
+	logger             logr.Logger
+	config             config.Config
+	annotationMutation *auto.AnnotationMutators
```
sky333999, line 39: nit: annotationMutation -> annotationMutator

internal/webhook/workloadmutation/webhookhandler.go
```diff
@@ -0,0 +1,98 @@
...
+func (p *workloadMutationWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
+	if p.annotationMutation == nil {
+		return admission.Errored(http.StatusBadRequest, errors.New("failed to unmarshal annotation config"))
+	}
+
+	var err error
+	var marshaledObject []byte
+	var object runtime.Object
+	switch objectKind := req.Kind.Kind; objectKind {
+	case "DaemonSet":
```
sky333999, line 62: Can we look for consts for these from the k8s lib?

internal/webhook/workloadmutation/webhookhandler.go
```diff
@@ -0,0 +1,98 @@
...
+		decoder:            decoder,
+		logger:             logger,
+		client:             cl,
+		annotationMutation: annotationMutation,
+	}
+}
+
+func (p *workloadMutationWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
+	if p.annotationMutation == nil {
+		return admission.Errored(http.StatusBadRequest, errors.New("failed to unmarshal annotation config"))
```
sky333999, line 55: Not a blocker, but can we manually test for this scenario? Given our ignore failurePolicy, want to confirm we still proceed with the creation of the resource.

internal/webhook/workloadmutation/webhookhandler.go
```diff
@@ -0,0 +1,98 @@
...
+		p.annotationMutation.MutateStatefulSet(&ss)
+		object = &ss
+	default:
+		return admission.Errored(http.StatusBadRequest, errors.New("failed to unmarshal request object"))
+	}
+
+	marshaledObject, err = json.Marshal(object)
+	if err != nil {
+		res := admission.Errored(http.StatusInternalServerError, err)
+		res.Allowed = true
```
sky333999, line 93: I see we set this here but not when returning the errors above. Does this change the behavior?

main.go
```diff
@@ -179,6 +180,26 @@ func main() {
 		os.Exit(1)
 	}
 
+	var autoAnnotationConfig auto.AnnotationConfig
+	var autoAnnotation *auto.AnnotationMutators
```
sky333999, line 184: nit: autoAnnotation -> autoAnnotationMutator

main.go
```diff
@@ -179,6 +180,26 @@ func main() {
...
+	if os.Getenv("DISABLE_AUTO_ANNOTATION") != "true" {
+		if err := json.Unmarshal([]byte(autoAnnotationConfigStr), &autoAnnotationConfig); err != nil {
+			setupLog.Error(err, "unable to unmarshal auto-annotation config")
+		} else {
+			autoAnnotation = auto.NewAnnotationMutators(
+				mgr.GetClient(),
+				mgr.GetAPIReader(),
```
sky333999, line 191: What does this get us? Is it to bypass the cache? Do we have a usecase for that?

pkg/instrumentation/auto/annotation.go
```diff
@@ -23,9 +23,12 @@ const (
 	defaultAnnotationValue = "true"
 )
 
+// +kubebuilder:rbac:groups="",resources=namespaces,verbs=list;watch;update
```
sky333999, line 26: We will have to create a namespace mutating webhook and can move this in there once thats ready.

main.go
```diff
@@ -179,6 +180,26 @@ func main() {
...
+				logger,
+				autoAnnotationConfig,
+				instrumentation.NewTypeSet(
+					instrumentation.TypeJava,
+					instrumentation.TypePython,
+				),
+			)
+			go autoAnnotation.MutateAll(ctx)
```
sky333999, line 199: I think we should call this after we register the webhook etc towards the end just before actually starting the manager to minimize race conditions.

internal/webhook/workloadmutation/webhookhandler.go
jefchien: I was looking for this too. Wasn't able to find it.

internal/webhook/workloadmutation/webhookhandler.go
jefchien: To be fair, this is the same behavior as the podmutation webhook if it fails to decode the pod.

Instead of doing this check here, can we just change main to not create the webhook? If AnnotationMutators is nil, it means the config is invalid or the auto annotations were disabled.

main.go
jefchien: It's to bypass the cache. During testing, we ran into a similar issue to https://github.com/jaegertracing/jaeger-operator/issues/809 where the cache wasn't available for reads yet.

internal/webhook/workloadmutation/webhookhandler.go
mitali-salvi: Couldnt find constants, we could declare our own

main.go
mitali-salvi: Even if we move it after the webhook, we will still run into the race condition, now the webhook will try to annotate first before the goroutine. We could add a delay but that's going to get tricky as well

I dont have strong thoughts about either

internal/webhook/workloadmutation/webhookhandler.go
mitali-salvi: I am going to move this logic to main.go since this block workload creation. I do have the option to allow it but I think moving the logic to main makes mroe sense

config/rbac/role.yaml
```diff
@@ -32,6 +32,7 @@ rules:
   - namespaces
   verbs:
   - list
+  - update
```
lisguo, line 35: can we get these changes in the helm chart too?

pkg/instrumentation/auto/annotation_test.go
```diff
@@ -106,6 +106,7 @@ func TestAnnotationMutators_Namespaces(t *testing.T) {
 			ctx := context.Background()
 			client := fake.NewClientBuilder().WithLists(&corev1.NamespaceList{Items: namespaces}).Build()
 			mutators := NewAnnotationMutators(
+				client,
```
lisguo, line 109: did you mean clientReader?

jefchien: The client reader is just an interface. We can reuse the fake client for both.

config/rbac/role.yaml
mitali-salvi: yeah I plan to create a separate PR to update the helm chart

main.go
sky333999: The webhook wont get triggered unless there was a create or update to a resource in the gap between the two. So i think thats far less likely to happen.
But agree we need something more deterministic like checking the manager is up and running via a health check or something.

main.go
```diff
@@ -179,6 +180,30 @@ func main() {
 		os.Exit(1)
 	}
 
+	decoder := admission.NewDecoder(mgr.GetScheme())
+
+	var autoAnnotationConfig auto.AnnotationConfig
+	var autoAnnotationMutator *auto.AnnotationMutators
```
jefchien, line 186: nit: These don't need to be declared outside of the block. They are not used anywhere else.

internal/webhook/workloadmutation/webhookhandler_test.go
```diff
@@ -0,0 +1,78 @@
...
+		name              string
+		expected          int32
+		allowed           bool
+	}{
+		{
+			name:              "invalid payload",
+			req:               admission.Request{},
+			expected:          http.StatusBadRequest,
+			allowed:           false,
+			autoAnnotationStr: "{\"java\":{\"namespaces\":[\"test\"],\"daemonsets\":[\"default/prometheus-daemonset\"]}}",
```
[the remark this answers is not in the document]
mitali-salvi: removed

main.go
mitali-salvi: updated

internal/webhook/workloadmutation/webhookhandler_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/aws/amazon-cloudwatch-agent-operator/pkg/instrumentation"
+	"github.com/aws/amazon-cloudwatch-agent-operator/pkg/instrumentation/auto"
+)
+
+var (
+	k8sClient client.Client
+	logger    = logf.Log.WithName("unit-tests")
+)
+
+func TestInvalidRequest(t *testing.T) {
```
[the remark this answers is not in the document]
mitali-salvi: updated

jefchien (approved):

internal/webhook/workloadmutation/webhookhandler.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+func (p *workloadMutationWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
+	var err error
+	var marshaledObject []byte
+	var object runtime.Object
+	switch objectKind := req.Kind.Kind; objectKind {
+	case "DaemonSet":
+		ds := appsv1.DaemonSet{}
```
okankoAMZ, line 59: nit: Can't we name `ds` or `d` or `ss` something more generic then move the Decode, error handling, object assigning   out of the switch statement ?

main.go
```diff
@@ -179,6 +180,29 @@ func main() {
 		os.Exit(1)
 	}
 
+	decoder := admission.NewDecoder(mgr.GetScheme())
+
+	if os.Getenv("DISABLE_AUTO_ANNOTATION") != "true" {
```
okankoAMZ, line 185: nit: Can we type cast these env vars as boolean, so that we do not check them against string versions of bools?

okankoAMZ, line 185: nit: If we define these env variables as global vars, it would make it easier in the long run to update them and re-use them. `"DISABLE_AUTO_ANNOTATION_KEY = "DISABLE_AUTO_ANNOTATION"`

main.go
```diff
@@ -179,6 +180,29 @@ func main() {
...
+					instrumentation.TypePython,
+				),
+			)
+			mgr.GetWebhookServer().Register("/mutate-v1-workload", &webhook.Admission{
+				Handler: workloadmutation.NewWebhookHandler(cfg, ctrl.Log.WithName("workload-webhook"), decoder, mgr.GetClient(), autoAnnotationMutator)})
+			go autoAnnotationMutator.MutateAll(ctx)
+		}
+	}
+
 	if os.Getenv("ENABLE_WEBHOOKS") != "false" {
```
okankoAMZ, line 206: same as my previous comment

okankoAMZ (approved):
No, blockers but commented couple nice to haves

sky333999 (approved):

PR #80, opened 2024-02-05
Logic to generate default cluster_name if not supplied

mitali-salvi:
*Description of changes:*
1. Implement logic to generate a default cluster-name (generated using the release name) if not supplied in Helm chart. This cluster name will be passed down to the cloudwatch-agent and fluent-bit config
2. Marking region as a required field. this region will be passed down to both the CWA and fluent-bit config

*Testing*
cluster_name is not provided, attaching snippets rather than the entire YAML
```
CWA --
  image: public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.300031.1b317
  mode: daemonset
  serviceAccount: cloudwatch-agent
  config: "{\"agent\":{\"region\":\"us-east-1\"},\"logs\":{\"metrics_collected\":{\"app_signals\":{\"hosted_in\":\"k8s-cluster-0ba3ba5\"},\"kubernetes\":{\"cluster_name\":\"k8s-cluster-0ba3ba5\",\"enhanced_container_insights\":true}}},\"traces\":{\"traces_collected\":{\"app_signals\":{}}}}"

FluentBit --
        image: public.ecr.aws/aws-observability/aws-for-fluent-bit:2.31.12.20230911
        imagePullPolicy: Always
        env:
        - name: AWS_REGION
          value: us-east-1
        - name: CLUSTER_NAME
          value: k8s-cluster-0ba3ba5
```

cluster_name is provided via the helm chart, attaching snippets rather than the entire YAML
```
helm upgrade --install  --debug --namespace amazon-cloudwatch amazon-cloudwatch-operator ./ --create-namespace --set clusterName=cloudwatchagent-operator --set region=us-east-1


CWA --
  image: public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.300031.1b317
  mode: daemonset
  serviceAccount: cloudwatch-agent
  config: "{\"logs\":{\"metrics_collected\":{\"app_signals\":{\"hosted_in\":\"k8s-cluster-0ba3ba5\"},\"kubernetes\":{\"cluster_name\":\"k8s-cluster-0ba3ba5\",\"enhanced_container_insights\":true}}},\"traces\":{\"traces_collected\":{\"app_signals\":{}}}}"

FluentBit --
        image: public.ecr.aws/aws-observability/aws-for-fluent-bit:2.31.12.20230911
        imagePullPolicy: Always
        env:
        - name: AWS_REGION
          value: us-east-1
        - name: CLUSTER_NAME
          value: "k8s-cluster-0ba3ba5"
```

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -116,7 +116,7 @@ Common labels
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
-app.kubernetes.io/managed-by: EKS
+app.kubernetes.io/managed-by: "AmazonCloudWatchAgentOperator"
 {{- end }}
 
 {{/*
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -22,6 +22,11 @@ Helper function to modify cloudwatch-agent config
 {{- define "cloudwatch-agent.config-modifier" -}}
 {{- $configCopy := deepCopy .Values.agent.config }}
 
+{{- $agent := pluck "agent" $configCopy | first }}
+{{- if and (empty $agent) (empty $agent.region) }}
+{{- $agent := set $agent "region" .Values.region }}
+{{- end }}
+
 {{- $appSignals := pluck "app_signals" $configCopy.logs.metrics_collected | first }}
 {{- if empty $appSignals.hosted_in }}
 {{- $appSignals := set $appSignals "hosted_in" (include "kubernetes-cluster.name" .) }}
@@ -51,6 +56,8 @@ Helper function to modify default agent config
 */}}
 {{- define "cloudwatch-agent.modify-default-config" -}}
 {{- $configCopy := deepCopy .Values.agent.defaultConfig }}
+{{- $agent := pluck "agent" $configCopy | first }}
+{{- $agent := set $agent "region" .Values.region }}
 {{- $appSignals := pluck "app_signals" $configCopy.logs.metrics_collected | first }}
 {{- $appSignals := set $appSignals "hosted_in" (include "kubernetes-cluster.name" .) }}
 {{- $containerInsights := pluck "kubernetes" $configCopy.logs.metrics_collected | first }}
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -56,8 +56,8 @@ Helper function to modify default agent config
 */}}
 {{- define "cloudwatch-agent.modify-default-config" -}}
 {{- $configCopy := deepCopy .Values.agent.defaultConfig }}
-{{- $agent := pluck "agent" $configCopy | first }}
-{{- $agent := set $agent "region" .Values.region }}
+{{- $agentRegion := dict "region" .Values.region }}
+{{- $agent := set $configCopy "agent" $agentRegion }}
 {{- $appSignals := pluck "app_signals" $configCopy.logs.metrics_collected | first }}
 {{- $appSignals := set $appSignals "hosted_in" (include "kubernetes-cluster.name" .) }}
 {{- $containerInsights := pluck "kubernetes" $configCopy.logs.metrics_collected | first }}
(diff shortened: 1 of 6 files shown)


helm/templates/_helpers.tpl
```diff
@@ -5,6 +5,59 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Name of k8s cluster
+*/}}
+{{- define "kubernetes-cluster.name" -}}
+{{- if eq .Values.clusterName "EKS_CLUSTER_NAME" }}
```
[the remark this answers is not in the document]
lisguo: I think the intent is that if it equals the hardcoded string "EKS_CLUSTER_NAME", then it is not on EKS. Therefore we can create a generated cluster name

helm/templates/_helpers.tpl
```diff
@@ -56,7 +116,7 @@ Common labels
 {{- if .Chart.AppVersion }}
 app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
-app.kubernetes.io/managed-by: EKS
+app.kubernetes.io/managed-by: "AmazonCloudWatchAgentOperator"
```
lisguo, line 119: nit: amazon-cloudwatch-agent-operator

helm/values.yaml
```diff
@@ -137,6 +137,8 @@ agent:
   config: # optional config that can be provided to override the defaultConfig
   defaultConfig:
     {
+      "agent": {
```
lisguo, line 140: nit: can prob remove the agent section entirely

lisguo (dismissed):

helm/templates/_helpers.tpl
mitali-salvi: updated

helm/values.yaml
mitali-salvi: removed and moved the logic to the helper function

lisguo (approved):

lisguo (approved):

movence (approved):

helm/templates/fluent-bit-daemonset.yaml
```diff
@@ -28,7 +31,7 @@ spec:
         - name: AWS_REGION
           value: {{ .Values.region }}
         - name: CLUSTER_NAME
-          value: {{ .Values.clusterName | quote }}
+          value: {{ include "kubernetes-cluster.name" . }}
```
sky333999, line 34: @movence doesnt this bring back the bug you fixed for numeric only cluster names not being quoted?

helm/templates/_helpers.tpl
```diff
@@ -5,6 +5,66 @@ Expand the name of the chart.
...
+{{- define "kubernetes-cluster.name" -}}
+{{- if empty .Values.clusterName }}
+{{- default "" (printf "k8s-cluster-%s" (sha256sum .Release.Name | trunc 7)) }}
+{{- else }}
+{{- default "" .Values.clusterName }}
+{{- end }}
+{{- end }}
+
+{{/*
+Helper function to modify cloudwatch-agent config
```
sky333999, line 20: Whats diff between the two funcs below? 
I would have expected a generic helper function that takes a config (either user provided or default) as an input and then just merges region & cluster name on top of that.
```
.Values.agent.config | injectRegionAndClusterName | toJson | quote
.Values.agent.defaultConfig | injectRegionAndClusterName | toJson | quote
```

helm/values.yaml
```diff
@@ -11,11 +11,11 @@ nameOverride: ""
...
 imagePullSecrets: [ ]
 
-## Provide the ClusterName (optional when installing via EKS add-on)
-clusterName: EKS_CLUSTER_NAME
+## Provide the ClusterName
+clusterName:
 
-## Provide the Region (optional when installing via EKS add-on)
-region: AWS_REGION_NAME
+## Provide the Region
```
sky333999, line 17: Can we keep the comments on these saying this is required and the cluster name is optional and we generate one if not provdied

helm/templates/_helpers.tpl
```diff
@@ -5,6 +5,66 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Name of k8s cluster
+*/}}
+{{- define "kubernetes-cluster.name" -}}
+{{- if empty .Values.clusterName }}
+{{- default "" (printf "k8s-cluster-%s" (sha256sum .Release.Name | trunc 7)) }}
```
sky333999, line 13: Is the release name unique enough for our purposes here? Is a hash good enough or do we need randomness?
If the same release is installed on multiple clusters with the same name, wont all the clusters have the same hash?

movence: This is a good point, and we should add some randomness to it.

PR #81, opened 2024-02-06
add dcgm-exporter daemonset for nvidia

movence:
*Description of changes:*
- Add new `dcgm-exporter` daemonset to support NVIDIA GPU metrics with k8s
- Add new service that uses `dcgm-exporter` pods where the traffic is limited within node (no cross nodes communication)

The new `dcgm-exporter` pod has `nodeAffinity` config that spins up itself in only GPU nodes. The supported GPU instances are hard coded into `values.yaml` which can be extended when there are more GPU instances/types supported. 

Below is the list of pods in test cluster where 1 GPU and 1 non-gpu nodes are. 
```
NAME                                                              READY   STATUS    RESTARTS   AGE
amazon-cloudwatch-observability-controller-manager-xxxxxx   1/1     Running   0          4m17s
cloudwatch-agent-xxxxxx                                            1/1     Running   0          4m16s
cloudwatch-agent-xxxxxx                                            1/1     Running   0          4m17s
dcgm-exporter-xxxxxx                                               1/1     Running   0          4m18s
fluent-bit-xxxxxx                                                 1/1     Running   0          4m18s
fluent-bit-xxxxxx                                                  1/1     Running   0          4m17s
```

### DIFF
--- a/helm/templates/dcgm-exporter-daemonset.yaml
+++ b/helm/templates/dcgm-exporter-daemonset.yaml
@@ -1,0 +1,71 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: dcgm-exporter
+  namespace: {{ .Release.Namespace }}
+  labels:
+    k8s-app: dcgm-exporter
+    version: v1
+spec:
+  selector:
+    matchLabels:
+      k8s-app: dcgm-exporter
+  template:
+    metadata:
+      labels:
+        k8s-app: dcgm-exporter
+        version: v1
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: {{ template "cloudwatch-agent.serviceAccountName" . }}
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+            - matchExpressions:
+              - key: {{ .Values.gpuNodeLabelKey }}
+                operator: In
+                values: {{ .Values.gpuInstances | toYaml | nindent 16 }}
+      containers:
+      - name: dcgm-exporter
+        securityContext:
+          privileged: true
+        image: "{{ .Values.dcgmExporter.image.repository }}:{{ .Values.dcgmExporter.image.tag }}"
+        args:
+        {{- range $.Values.dcgmExporter.arguments }}
+        - {{ . }}
+        {{- end }}
+        env:
+        - name: "DCGM_EXPORTER_KUBERNETES"
+          value: "true"
+        - name: "DCGM_EXPORTER_LISTEN"
+          value: "{{ .Values.dcgmExporter.service.address }}"
+        - name: "DCGM_EXPORTER_COLLECTORS"
+          value: "/etc/dcgm-exporter/dcp-metrics-included.csv"
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        ports:
+        - name: "metrics"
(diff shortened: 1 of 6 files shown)


helm/templates/dcgm-exporter-daemonset.yaml
```diff
@@ -0,0 +1,71 @@
...
+    matchLabels:
+      k8s-app: dcgm-exporter
+  template:
+    metadata:
+      labels:
+        k8s-app: dcgm-exporter
+        version: v1
+    spec:
+      priorityClassName: system-node-critical
+      serviceAccountName: {{ template "cloudwatch-agent.serviceAccountName" . }}
```
lisguo, line 20: should we make a different service account for the dcgm exporter instead of using the cloudwatch agent service account? There might be different privileges

helm/values.yaml
```diff
@@ -35,10 +39,12 @@ containerLogs:
 manager:
   name:
   image:
-    repository: cloudwatch-agent-operator
+    # repository: cloudwatch-agent-operator
```
lisguo, line 42: probably need to revert to `cloudwatch-agent-operator`

helm/values.yaml
```diff
@@ -35,10 +39,12 @@ containerLogs:
...
+    repository: cwagent-operator-release
     tag: 1.0.2
     repositoryDomainMap:
-      public: public.ecr.aws/cloudwatch-agent
+      # public: public.ecr.aws/cloudwatch-agent
```
lisguo, line 46: probably need to revert this

helm/values.yaml
```diff
@@ -123,10 +129,13 @@ admissionWebhooks:
 agent:
   name:
   image:
-    repository: cloudwatch-agent
-    tag: 1.300031.1b317
+    # repository: cloudwatch-agent
+    # tag: 1.300031.1b317
+    repository: tupperware
```
lisguo, line 134: Are there changes to the agent that we need to make to get this to work?

helm/values.yaml
```diff
@@ -137,6 +146,9 @@ agent:
   config: # optional config that can be provided to override the defaultConfig
   defaultConfig:
     {
+      "agent": {
+        "debug": true
```
lisguo, line 150: remove

helm/values.yaml
```diff
@@ -150,4 +162,32 @@ agent:
...
       }
-    }
\ No newline at end of file
+    }
+  enableGpuConfig: 
+    {
+      "logs": {
+        "metrics_collected": {
+          "kubernetes": {
+            "gpu_metrics": true
```
lisguo, line 171: instead of having this as true, can we just have it an empty map in case we want to add more config options in the future?

helm/values.yaml
```diff
@@ -150,4 +162,32 @@ agent:
...
+          },
+        }
+      }
+    }
+
+dcgmExporter:
+  name:
+  image:
+    repository: nvcr.io/nvidia/k8s/dcgm-exporter
```
lisguo, line 180: I am assuming this will be our mirrored repo eventually

movence: correct

helm/values.yaml
movence: removed. it was a leftover from the previous revision where there were another agent daemonset specifically for GPU nodes.

helm/values.yaml
movence: nope. removed test repo

helm/templates/dcgm-exporter-daemonset.yaml
movence: Yeah, great call. updated

PR #83, opened 2024-02-07
Add namespace webhook for annotations

jefchien:
*Issue #, if available:* N/A

*Description of changes:* Adds a webhook to handle creates/updates to namespaces. If a namespace that is configured for auto-annotation is created or modified, the webhook will attempt to update the annotations on it.

*Testing:*
Set the auto-annotation configuration.
```
> kubectl set env deployment/cloudwatch-controller-manager AUTO_ANNOTATION_CONFIG="{\"java\":{\"namespaces\":[\"coffee-shop\"]},\"python\":{\"namespaces\":[\"snake-skin\"]}}" -n amazon-cloudwatch
deployment.apps/cloudwatch-controller-manager env updated
```
Created namespace.
```
> kubectl create ns snake-skin
namespace/snake-skin created
```
Namespace was annotated.
```
kubectl describe ns snake-skin
Name:         snake-skin
Labels:       kubernetes.io/metadata.name=snake-skin
Annotations:  cloudwatch.aws.amazon.com/auto-annotate-python: true
              instrumentation.opentelemetry.io/inject-python: true
Status:       Active

No resource quota.

No LimitRange resource.
```

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -70,6 +70,26 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
+      path: /mutate-v1-namespace
+  failurePolicy: Ignore
+  name: mnamespace.kb.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - namespaces
+  sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
       path: /mutate-v1-workload
   failurePolicy: Ignore
   name: mworkload.kb.io
--- a/internal/webhook/namespacemutation/webhookhandler.go
+++ b/internal/webhook/namespacemutation/webhookhandler.go
@@ -1,0 +1,47 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package namespacemutation
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	"github.com/aws/amazon-cloudwatch-agent-operator/pkg/instrumentation/auto"
+)
+
+// +kubebuilder:webhook:path=/mutate-v1-namespace,mutating=true,failurePolicy=ignore,groups="",resources=namespaces,verbs=create;update,versions=v1,name=mnamespace.kb.io,sideEffects=none,admissionReviewVersions=v1
+
+var _ admission.Handler = (*handler)(nil)
+
+type handler struct {
+	decoder            *admission.Decoder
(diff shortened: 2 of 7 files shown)


internal/webhook/namespacemutation/webhookhandler.go
```diff
@@ -0,0 +1,47 @@
...
+	namespace := &corev1.Namespace{}
+	err := h.decoder.Decode(req, namespace)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+	h.annotationMutators.MutateObject(namespace)
+	marshaledNamespace, err := json.Marshal(namespace)
+	if err != nil {
+		res := admission.Errored(http.StatusInternalServerError, err)
+		res.Allowed = true
```
okankoAMZ, line 43: Is there a specific reason why we are setting setting Allowed for the response there?

main.go
```diff
@@ -11,6 +11,8 @@ import (
 	"runtime"
 	"strings"
 
+	"github.com/aws/amazon-cloudwatch-agent-operator/internal/webhook/namespacemutation"
+
```
okankoAMZ, line 15: nit: this extra space might trigger the linter

okankoAMZ (dismissed):

config/webhook/manifests.yaml
```diff
@@ -64,6 +64,26 @@ webhooks:
     resources:
     - pods
   sideEffects: None
+- admissionReviewVersions:
```
lisguo, line 67: Can we get this added to the helm chart as well? https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/helm/templates/admission-webhooks/operator-webhook.yaml#L84

internal/webhook/namespacemutation/webhookhandler.go
lisguo: Related Docs: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#response

Docs specify that all responses must have an `allowed` setting. I guess in this case, the request is allowed but there was an error with marshalling the namespace.

lisguo: On a successful patch, `PatchResponseFromRaw` actually returns Allowed to true.

https://github.com/kubernetes-sigs/controller-runtime/blob/v0.17.0/pkg/webhook/admission/response.go#L88

lisguo (approved):

internal/webhook/namespacemutation/webhookhandler.go
lisguo: Interesting, the `Errored` func has Allowed to false which I presume we don't want.

config/webhook/manifests.yaml
jefchien: Can make the changes to the helm chart in a separate PR since we'll need to add the workload webhook as well.

internal/webhook/namespacemutation/webhookhandler.go
jefchien: Yeah at this point it's a valid namespace, so even if we want to report an error, we still want to mark it as a valid request.

okankoAMZ (approved):

PR #85, opened 2024-02-09
Helm updates to add autoAnnotationConfig and Namespace/Workload Webhooks

sky333999:
*Issue #, if available: N/A

*Description of changes:* Updates to helm chart:
* Add mutating webhook definitions for namespace and workload mutators
* Add autoAnnotation config and pass it to the controller manager
* Update RBAC permissions to allow mutating namespaces

### DIFF
--- a/helm/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
+++ b/helm/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
@@ -95,6 +95,66 @@ webhooks:
     - pods
   sideEffects: None
   timeoutSeconds: {{ .Values.admissionWebhooks.timeoutSeconds }}
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: {{ template "amazon-cloudwatch-observability.webhookServiceName" . }}
+      namespace: {{ .Release.Namespace }}
+      path: /mutate-v1-namespace
+  failurePolicy: {{ .Values.admissionWebhooks.pods.failurePolicy }}
+  name: mnamespace.kb.io
+    {{- if .Values.admissionWebhooks.namespaceSelector }}
+  namespaceSelector:
+    {{- toYaml .Values.admissionWebhooks.namespaceSelector | nindent 6 }}
+    {{- end }}
+    {{- if .Values.admissionWebhooks.objectSelector }}
+  objectSelector:
+    {{- toYaml .Values.admissionWebhooks.objectSelector | nindent 6 }}
+    {{- end }}
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - namespaces
+  sideEffects: None
+  timeoutSeconds: {{ .Values.admissionWebhooks.timeoutSeconds }}
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: {{ template "amazon-cloudwatch-observability.webhookServiceName" . }}
+      namespace: {{ .Release.Namespace }}
+      path: /mutate-v1-workload
+  failurePolicy: {{ .Values.admissionWebhooks.pods.failurePolicy }}
+  name: mworkload.kb.io
+    {{- if .Values.admissionWebhooks.namespaceSelector }}
+  namespaceSelector:
+    {{- toYaml .Values.admissionWebhooks.namespaceSelector | nindent 6 }}
+    {{- end }}
+    {{- if .Values.admissionWebhooks.objectSelector }}
+  objectSelector:
+    {{- toYaml .Values.admissionWebhooks.objectSelector | nindent 6 }}
+    {{- end }}
+  rules:
+  - apiGroups:
+    - apps
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    - UPDATE
+    resources:
+    - daemonsets
+    - deployments
+    - statefulsets
+  sideEffects: None
+  timeoutSeconds: {{ .Values.admissionWebhooks.timeoutSeconds }}
 ---
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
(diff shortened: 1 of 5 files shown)


helm/templates/operator-deployment.yaml
```diff
@@ -26,6 +26,7 @@ spec:
       containers:
       - image: {{ template "cloudwatch-agent-operator.image" . }}
         args:
+        - {{ printf "--auto-annotation-config=%s" (.Values.manager.autoAnnotation | toJson) | quote }}
```
sky333999, line 29: I spent a lot of time trying to understand why the following wouldnt work:
```
- "--auto-annotation-config={{ (.Values.manager.autoAnnotation | toJson | quote) }}"
```
It fails with an error saying `Error: YAML parse error on amazon-cloudwatch-observability/templates/operator-deployment.yaml: error converting YAML to JSON: yaml: line 29: did not find expected key`

After a lot of trial & error, the above is what finally worked to achieve the same end goal.

helm/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
```diff
@@ -95,6 +95,66 @@ webhooks:
...
+  namespaceSelector:
+    {{- toYaml .Values.admissionWebhooks.namespaceSelector | nindent 6 }}
+    {{- end }}
+    {{- if .Values.admissionWebhooks.objectSelector }}
+  objectSelector:
+    {{- toYaml .Values.admissionWebhooks.objectSelector | nindent 6 }}
+    {{- end }}
+  rules:
+  - apiGroups:
+    - ""
```
straussb, line 117: `"core"`? ([ref](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.29/#namespace-v1-core))

helm/values.yaml
```diff
@@ -50,6 +50,17 @@ manager:
     python:
       repository: ghcr.io/open-telemetry/opentelemetry-operator/autoinstrumentation-python
       tag: 0.43b0
+  autoAnnotation:
```
straussb, line 53: `autoAnnotateAutoInstrumentation`?  I'm fine either way, just another possibility for you to think about...

helm/templates/operator-deployment.yaml
straussb: Does `quote` add double-quotes which are interfering with the literal double quotes you have here?

Is there a way to have Helm spit out the raw text that it generates after processing?

helm/values.yaml
sky333999: Yeah I was debating the same. Wasnt sure if thats too verbose.. but maybe that is more explicit & clear.

helm/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
sky333999: I picked this up from a manifest that @jefchien had tested in an earlier PR.
But I also just checked on a cluster and namespaces do not actually have an apiGroup. Im guessing empty defaults to core.
```
kubectl api-resources -o wide | grep "deployments\|namespaces"
namespaces                        ns                       v1                                     false        Namespace                        create,delete,get,list,patch,update,watch                    
deployments                       deploy                   apps/v1                                true         Deployment                       create,delete,deletecollection,get,list,patch,update,watch   all
```

helm/values.yaml
straussb: It is definitely verbose - but nobody is typing it in by hand :)

helm/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
sky333999: Right I see the same being referenced in a couple of places - that "" means `core`.

Also, in our `ClusterRole` [here](https://github.com/aws/amazon-cloudwatch-agent-operator/pull/85/files#diff-9cf6c12a0631332ed0ce9158f1c5c9b6fe11b36206c87e73f3347309a827d866R12) or in CI samples [here](https://github.com/aws-samples/amazon-cloudwatch-container-insights/blob/main/k8s-deployment-manifest-templates/deployment-mode/daemonset/container-insights-monitoring/cwagent/cwagent-serviceaccount.yaml#L14), we dont see to use `core`.

It shouldnt hurt to explicitly say `core`, so I dont mind testing and updating accordingly.

helm/templates/operator-deployment.yaml
sky333999: In the actual deployment manifest, it looks like follows:
```
- "--auto-annotation-config={\"java\":{\"daemonsets\":[],\"deployments\":[],\"namespaces\":[],\"statefulsets\":[]},\"python\":{\"daemonsets\":[],\"deployments\":[],\"namespaces\":[],\"statefulsets\":[]}}"
```

helm/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
sky333999: I tested this real quick and if I set this apiGroup to `core`, the webhook actually doesnt work as expected.

helm/values.yaml
```diff
@@ -50,6 +50,17 @@ manager:
...
+    java:
+      namespaces: [ ]
+      deployments: [ ]
+      daemonsets: [ ]
+      statefulsets: [ ]
+    python:
+      namespaces: [ ]
+      deployments: [ ]
+      daemonsets: [ ]
+      statefulsets: [ ]
```
jefchien, line 63: nit: Do we need to define these explicitly?

jefchien (approved):

helm/templates/admission-webhooks/operator-webhook-with-cert-manager.yaml
straussb: Thanks for checking!  Seems like a problem that there's no way to specify core only, but oh well, in practice it's not really an issue since we restrict to the namespaces resource.

straussb (approved):

helm/values.yaml
sky333999: I just kept these as a sort of self-documentation. If someone is just looking at our helm chart values.yaml, this makes it clear as to what options can be passed in.

PR #95, opened 2024-02-21
Modifying Operator test suites to accout for Python

Paramadon:
*Issue #, if available:*
None

*Description of changes:*

### Update Operator Integration Tests to account for Python

This pull request introduces a update to our operator testing framework through the integration of Python instrumentation. The primary goal of this update is to enable customers to use Python, thereby providing a means to verify their environment variables and ensure that their setup is correct. 


helm/values.yaml
```diff
@@ -12,10 +12,10 @@ nameOverride: ""
 imagePullSecrets: [ ]
 
 ## Provide the ClusterName
-clusterName:
+clusterName: NewClusterForPulse
 
 ## Provide the Region
-region:
+region: us-west-2
```
okankoAMZ, line 18: why do we have to define this region ?

helm/values.yaml
```diff
@@ -12,10 +12,10 @@ nameOverride: ""
 imagePullSecrets: [ ]
 
 ## Provide the ClusterName
-clusterName:
+clusterName: NewClusterForPulse
```
okankoAMZ, line 15: Do we want to predefined this ?

okankoAMZ (commented):
Left some comments

.github/workflows/operator-integration-test.yml
```diff
@@ -68,17 +68,42 @@ jobs:
...
+          kubectl rollout restart deployment nginx
+          sleep 60
+          kubectl get pods -A
+          kubectl describe pods -n default
+          cd ./cmd
+          go run validate_instrumentation_vars.go default ns_instrumentation_env_variables.json     
+
+      - name: Test for default instrumentation resources for python
+        run: |
+          cd integration-tests/manifests
```
lisguo, line 95: nit: I prefer to use absolute paths. So can just do `kubectl apply -f integration-tests/manifests/sample-deployment-python.yaml`

I know this was carry-over though

application-signals-demo
lisguo, line 1: Is there a reason this was added?

helm/values.yaml
lisguo, line 1: We shouldn't include this. This was just for manual testing setup

integration-tests/manifests/cmd/default_instrumentation_python_env_variables.json
lisguo, line 1: I wonder if it makes sense to have a separate folder under `integration-tests` for each language and revamp our directory structure? Maybe we can do something like
```
integration-tests
 |- java
   |- default_instrumentation_java_env_variables.json
   |- sample-deployment-java.yaml
 |- python
   |- default_instrumentation_python_env_variables.json
   |- sample-deployment-python.yaml
```

lisguo (changes requested):

application-signals-demo
Paramadon: removed

.github/workflows/operator-integration-test.yml
```diff
@@ -68,24 +68,43 @@ jobs:
...
-          kubectl apply -f ./sample-instrumentation.yaml
-          sleep 60
+          kubectl apply -f integration-tests/manifests/sample-instrumentation.yaml
+          sleep 5
           kubectl rollout restart deployment nginx
+          sleep 5
+          kubectl get pods -A
+          kubectl describe pods -n default
+          cd integration-tests/manifests/cmd
+          go run validate_instrumentation_vars.go default ns_instrumentation_env_variables.json
```
lisguo, line 88: after we validate the namespaced instrumentation resource we should probably delete it via `kubectl delete instrumentation sample-instrumentation`

I'm surprised this didn't cause the python test to fail...since we would end up using the namespaced instrumentation instead of the defaults.

integration-tests/python/default_instrumentation_python_env_variables.json
```diff
@@ -0,0 +1,13 @@
+
+{
+  "PYTHONPATH":                              "/otel-auto-instrumentation-python/opentelemetry/instrumentation/auto_instrumentation:/otel-auto-instrumentation-python",
```
lisguo, line 3: these values are incorrect, it should be the env vars defined here: https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/pkg/instrumentation/defaultinstrumentation.go#L91

I think these are the namespaced ones, not the default instrumentation vars

.github/workflows/operator-integration-test.yml
lisguo: I see, the default python values are incorrect which is why it passed

.github/workflows/operator-integration-test.yml
```diff
@@ -68,24 +68,43 @@ jobs:
...
+          sleep 5
+          kubectl get pods -A
+          kubectl describe pods -n default
+          cd integration-tests/manifests/cmd
+          go run validate_instrumentation_vars.go default ns_instrumentation_env_variables.json     
+
+      - name: Test for default instrumentation resources for python
+        run: |
+          kubectl apply -f integration-tests/python/sample-deployment-python.yaml
           sleep 60
```
lisguo, line 93:
```suggestion
          sleep 5
```

.github/workflows/operator-integration-test.yml
```diff
@@ -68,24 +68,43 @@ jobs:
             echo "Pod is created test case: PASS"
           fi
 
-      - name: Test for default instrumentation resources
+      - name: Test for default instrumentation resources for Java
         run: |
-          cd integration-tests/manifests
-          kubectl apply -f ./sample-deployment.yaml
+          kubectl apply -f integration-tests/java/sample-deployment-java.yaml
           sleep 60
```
lisguo, line 74:
```suggestion
          sleep 5
```
prob can be consistent with the 5s

lisguo (changes requested):

okankoAMZ (approved):

PR #97, opened 2024-02-22
Add todo's to fix CW agent container image

KlwntSingh:
*Description of changes:*
1. Added todo's to fix link to CI
2. Change tag to fluentbit on Windows

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -23,7 +23,7 @@ containerLogs:
     image:
       repository: aws-for-fluent-bit
       tag: 2.31.12.20230911
-      tagWindows: windowsservercore-latest
+      tagWindows: 2.31.12-windowsservercore
       repositoryDomainMap:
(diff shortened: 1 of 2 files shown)


chadpatel (approved):

sky333999 (approved):

helm/values.yaml
```diff
@@ -23,7 +23,7 @@ containerLogs:
     image:
       repository: aws-for-fluent-bit
       tag: 2.31.12.20230911
-      tagWindows: windowsservercore-latest
+      tagWindows: 2.31.12-windowsservercore
```
sky333999, line 26: is this temporary? i imagine we will use the same tag for linux and windows eventually with them being multi arch

KlwntSingh: This is not temporary.
ECR repo is same but tag for Windows is different from Linux container images. We will need two tags. We will need to keep values for both tags in sync over time.

PR #106, opened 2024-03-05
Modifying operator test for python and java simultaneous instrumentation

Paramadon:
*Issue #, if available:*
We have tests that account for python and java individually, but not for when they are both injected into a pod simultaneously. Can use validation method to verify env variables: https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/integration-tests/manifests/cmd/validate_instrumentation_vars.go.

*Description of changes:*
Modified operator test to account for python and java running simultaneously. 

AC:
1. Verify environment variables are generated for both python and java
2. Validated the environment variables for accuracy.


sethAmazon (approved):

.github/workflows/operator-integration-test.yml
```diff
@@ -111,8 +111,27 @@ jobs:
           go run validate_instrumentation_vars.go default ns_instrumentation_env_variables.json
           kubectl delete instrumentation sample-instrumentation
 
+      - name: Test for default instrumentation resources for python and java
+        run: |
+          kubectl apply -f integration-tests/python-java/sample-deployment-python-java.yaml
+          sleep 5
```
lisguo, line 117: instead of sleep, we can use `kubectl wait`

https://kubernetes.io/docs/reference/kubectl/generated/kubectl_wait/

can do something like

```
kubectl wait --for=condition=Ready pod --all -n default
```

the above will wait for all pods in the default namespace to be ready

.github/workflows/operator-integration-test.yml
```diff
@@ -111,8 +111,27 @@ jobs:
...
+          kubectl describe pods -n default
+          kubectl describe pods -n amazon-cloudwatch
+          go run integration-tests/manifests/cmd/validate_instrumentation_vars.go default integration-tests/python-java/default_instrumentation_python-java_env_variables.json
+
 
 
+      - name: Test for defined instrumentation resources for python and java
+        run: |
+          kubectl apply -f integration-tests/manifests/sample-instrumentation.yaml
+          sleep 5
```
lisguo, line 128: don't we need to restart our pods to check for the new instrumentation values? something like

```
kubectl delete pods --all -n default
kubectl wait --for=condition=Ready pod --all -n default
```

this will delete pods in the default namespace and waits for it to be ready again

.github/workflows/operator-integration-test.yml
```diff
@@ -71,47 +72,76 @@ jobs:
...
+
 
 
+      - name: Test for defined instrumentation resources for python and java
+        run: |
+          kubectl apply -f integration-tests/manifests/sample-instrumentation.yaml
+          kubectl delete pods --all -n default
+          sleep 3
+          kubectl wait --for=condition=Ready pod --all -n default
+          kubectl rollout restart deployment nginx
```
lisguo, line 137: why are we deleting pods and then doing a rollout restart anyway? We should probably choose 1.

Maybe just do the kubectl rollout and then follow up with kubectl wait

.github/workflows/operator-integration-test.yml
```diff
@@ -157,8 +182,6 @@ jobs:
           sleep 5
           go test -v -run TestJavaOnlyDeployment ./integration-tests/manifests/annotations -timeout 30m
           sleep 5
-          go test -v -run TestPythonOnlyDeployment ./integration-tests/manifests/annotations -timeout 30m
```
lisguo, line 160: we are removing the python only test?

lisguo (approved):

PR #129, opened 2024-03-15
Add python E2E test

zzhlogin:
This PR add python E2E test with feature parity as [Java E2E test](https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/.github/workflows/appsignals-e2e-test.yml).

An example test workflow run: https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/8365537415/job/22903829040


srprash (approved):
Small comment. Otherwise LGTM.

.github/workflows/appsignals-python-e2e-test.yml
```diff
@@ -0,0 +1,321 @@
+# This is a reusable workflow for running the Python E2E test for App Signals.
+# It is meant to be called from another workflow.
+# This E2E test is responsible for validating setting up a sample application on an EKS cluster and enabling
+# App Signals using the staging image of the CloudWatch Agent Operator. It validates the generated telemetry
+# including logs, metrics, and traces, then cleans up the cluster. The testing resources can be found in the
+# ADOT java instrumentation repo: https://github.com/aws-observability/aws-otel-java-instrumentation/tree/main/testing
```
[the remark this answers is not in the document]
zzhlogin: Updated

.github/workflows/build-and-upload-staging.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           tags: ${{ env.ECR_OPERATOR_STAGING_IMAGE }}
           platforms: linux/amd64, linux/arm64
 
-  e2e-test:
+  java-e2e-test:
```
lisguo, line 61: can we have a single e2e workflow that has both java and python? We are using our e2e testing as part of our release workflow as well which will make it easier for us to run our whole e2e suite

.github/workflows/appsignals-python-e2e-test.yml
```diff
@@ -0,0 +1,321 @@
...
+jobs:
+  appsignals-e2e-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          # TODO: This step is used for custom pre-release instrumentation
+          # Uisng SHA as of March 4. It is a temporary measure until the cw add-on is released with python support
+          ref: abf75babe672412cb63c56cbcf1c5ce2d8c97a1c
```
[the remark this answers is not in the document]
zzhlogin: Yes, the difference it the helm change. I tested helm using main branch, and the CW manager log shows:
```
kubectl logs -f amazon-cloudwatch-observability-controller-manager-867b6b8fztzf -n amazon-cloudwatch
{"level":"info","ts":"2024-03-15T21:13:21Z","msg":"Starting the Amazon CloudWatch Agent Operator","amazon-cloudwatch-agent-operator":"","cloudwatch-agent":"public.ecr.aws/cloudwatch-agent/cloudwatch-agent:0.0.0","auto-instrumentation-java":"%!s(<nil>)/%!s(<nil>):%!s(<nil>)","auto-instrumentation-python":"%!s(<nil>)/%!s(<nil>):%!s(<nil>)","build-date":"","go-version":"go1.20.14","go-arch":"amd64","go-os":"linux"}
```
Then, when I restart my sample app pods, I get:
```
django-frontend-service-58c68f4998-clbvp     0/1     Init:InvalidImageName   0          79s
image-backend-6bd56fd569-2lmdg               0/1     Init:InvalidImageName   0          79s
vehicle-inventory-backend-59db98c8d8-dlc92   0/1     Init:InvalidImageName   0          79s
```
Seems both adot Java and python images values are  nil. Does this mean the operator doesn’t find the correct image, and failed with injection?

Also, I think I need to use staging image, since the test will be added into staging artifacts build.

.github/workflows/build-and-upload-staging.yml
zzhlogin: Yes, this is a choice too. We can plan on merging the two tests into one workflow. But one concern is if we want to monitor the java & python test separately,  which might be more convenience for problem detection and debugging?

zzhlogin: Updated to use a new workflow calling the both workflows.

lisguo (dismissed):

srprash (approved):

lisguo (approved):

sethAmazon:
The test files look very similar. Can they be combined into a single file that variables handle the differences?

zzhlogin:
> The test files look very similar. Can they be combined into a single file that variables handle the differences?

They are similar, but there is several difference between python and Java tests:
1. They use different terraform setup as we have separate terraform folders for java & python separately, the input parameters are different:https://github.com/aws-observability/aws-application-signals-test-framework/tree/main/terraform
2. The valiadator is different, we also separate valiadtor template for java & python separately: https://github.com/aws-observability/aws-application-signals-test-framework/tree/main/validator/src/main/resources/expected-data-template

The only part that Python and Java share the same steps are setting up CW operator.

It is possible to combined into a single file, but one concern is: if we want to monitor the java & python test separately,  which might be more convenience for problem detection and debugging? And in the future, as we add more support for more languages, the single workflow will seems to be overload. Probably make more sense to separate them?

sethAmazon (approved):

PR #131, opened 2024-03-18
Enablement Experience Integration Test

Paramadon:
*Issue #, if available:*
Currently there is nothing to test if pods have the correct annotations from the enablement experience especially for different use cases that could occur when users have these injections. If not tested correctly, users environment will not work as expected. 

*Description of changes:*
Added integration tests for enablement experience. 

## Deployments Cases
Use Case 1: Java and Python on Deployment - Both Python and Java should be annotated
Use Case 2: Java on Deployment and Python Should be Removed since both are injected on the deployment
Use Case 3: Python on Deployment and Java annotations should be removed from deployment

## Daemonset Cases
Use Case 4: Python and Java on DaemonSet - Both Python and Java should be on the Daemonset
Use Case 5: Java on DaemonSet and Python should be removed
Use Case 6: Python on DaemonSet Java annotation should be removed

## Namespace Cases
Use Case 7: Both Java and Python on Namespace - All pods in the name space should have both Python and Java annotations
Use Case 8: Java on Namespace and Python should be removed 
Use Case 9: Python on Namespace and Java annotation should be removed

## Statefulset Cases

Use Case 10: Python and Java on Stateful set 
Use Case 11: Java on Stateful set and Python should be removed
Use Case 12: Python on Stateful set and java should be removed


lisguo (changes requested):

integration-tests/manifests/annotations/validate_annotations.go
```diff
@@ -0,0 +1,823 @@
...
+	}
+	deployment, err = clientSet.AppsV1().Deployments("amazon-cloudwatch").Get(context.TODO(), "cloudwatch-controller-manager", metav1.GetOptions{})
+	if err != nil {
+		fmt.Printf("Error getting deployment: %v\n\n", err)
+		os.Exit(1)
+	}
+	//finding where index of --auto-annotation-config= is (if it doesn't exist it will be appended)
+	fmt.Println(indexOfAutoAnnotationConfigString, string(jsonStr))
+
+	updateAnnotationConfig(indexOfAutoAnnotationConfigString, deployment, string(jsonStr))
```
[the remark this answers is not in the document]
Paramadon: Done, but instead of just returning an error, I add the auto annotation if it doesn't exist (which it should but just in the worst case)

integration-tests/manifests/sample-daemonset.yaml
```diff
@@ -0,0 +1,25 @@
...
+    matchLabels:
+      name: fluent-bit
+  template:
+    metadata:
+      labels:
+        name: fluent-bit
+    spec:
+      containers:
+        - name: fluent-bit
+          image: nginx:latest
```
sky333999, line 18: Why does a fluent-bit container run an nginx image?

integration-tests/manifests/sample-deployment.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: apps/v1
+kind: Deployment
```
sky333999, line 2: I would use the exact same resource spec including the same image for all 3, i.e. Deployment, Daemonset and StatefulSet. Just change the Kind and keep them as simple as we can.
We can just call them sample-deployment, sample-daemonset, sample-statefulset etc.

integration-tests/manifests/sample-statefulset.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: my-statefulset
```
sky333999, line 4: Can we create a sample-namespace file as well and keep these resources in that namespace so they dont just end up in default?
We need to test scenarios that auto annotate a whole namespace as well.

integration-tests/manifests/annotations/validate_annotations.go
```diff
@@ -0,0 +1,752 @@
...
+func verifyAutoAnnotation(clientSet *kubernetes.Clientset) bool {
+
+	//updating operator deployment
+	deployment, err := clientSet.AppsV1().Deployments(amazonCloudwatchNamespace).Get(context.TODO(), amazonControllerManager, metav1.GetOptions{})
+	if err != nil {
+		fmt.Printf("Error getting deployment: %v\n\n", err)
+		os.Exit(1)
+	}
+
+	//---------------------------USE CASE 1 (Java and Python on Deployment) ----------------------------------------------
```
sky333999, line 67: We should avoid the behavior of one test bleeding into another.
We can make these all separate test cases that can then be run in parallel - just make sure each test case uses a dynamic namespace spun up for that particular test case so there is very little overlap.
Spinning up a new cluster per test case might be an overkill.

Paramadon: Question: How do run the tests in parallel? Are we changing all the test to run concurrently through threads and locks?

Paramadon: The reason why I did the bleed in behavior was mostly to test if there is a language already annotated but the user removes that language's annotation, than we should verify it removal.

integration-tests/manifests/sample-statefulset.yaml
Paramadon: We are testing the default namespace, but I can create sample-namespace, but you want me to put all the other sample apps in this one new namespace?

integration-tests/manifests/sample-daemonset.yaml
Paramadon: I thought we needed the same image for all the sample-apps which is why I did this.

sky333999:
I think the use cases we have are a good start, but I think we are still missing a few to give us confidence.
We can get a framework ready for the initial commit and then incrementally add more use cases.

Couple of the use cases that I see are missing:
* If a namespace is annotated, confirm that only apps that are not already annotated get restarted and now have annotations. i.e. Existing annotated apps should NOT be restarted.
* If a resource is already annotated, what is the behavior of adding it in the operator args as well?
* If a resource is manually annotated and now its namespace is added for auto annotation, the resource should not be modified and should not be restarted.
* If a resource that is auto annotated is manually patched to remove the annotations, our mutator should add back the annotations.

.github/workflows/operator-integration-test.yml
```diff
@@ -110,3 +110,18 @@ jobs:
...
           go run validate_instrumentation_vars.go default ns_instrumentation_env_variables.json
           kubectl delete instrumentation sample-instrumentation
+
+      - name: Test Annotations
+        run: |
+          kubectl apply -f integration-tests/manifests/sample-deployment.yaml
+          kubectl apply -f integration-tests/manifests/sample-daemonset.yaml
+          kubectl apply -f integration-tests/manifests/sample-statefulset.yaml
+          sleep 5
+          kubectl rollout restart deployment nginx
```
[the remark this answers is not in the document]
Paramadon: We do not, that'll be removed. Good catch!

.github/workflows/operator-integration-test.yml
```diff
@@ -110,3 +110,18 @@ jobs:
...
+          sleep 5
+          kubectl get pods -A
+          kubectl describe pods -n default
+          go run integration-tests/manifests/annotations/validate_annotations.go
```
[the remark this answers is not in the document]
Paramadon: We can, but I was just copying the format for the test we did for eks_resources

integration-tests/manifests/sample-statefulset.yaml
sky333999: Left another comment - I think it would be better to have a sample-namespace for some namespace related usecases but for most others, create a `sample-namespace-<hash>` on the fly so we isolate the test cases from one another.

Be sure to clean up resources after each test case i.e. delete the namespace once its done.

integration-tests/manifests/annotations/validate_annotations.go
sky333999: >The reason why I did the bleed in behavior was mostly to test if there is a language already annotated but the user removes that language's annotation, than we should verify it removal.

Yes that should be the behavior for one of the tests. But once that flow of events is done, it shouldnt impact any of the other tests.

sky333999: > How do run the tests in parallel?

You can look at `t.Parallel()`.

integration-tests/manifests/sample-statefulset.yaml
Paramadon: Okay will do, also we don't need a whole different operator manager for each right? I just put a lock around when the operator is used

jefchien:
> * If a resource is manually annotated and now its namespace is added for auto annotation, the resource should not be modified and should not be restarted.

is not the current behavior, but if we think it should be, then it's worth adding it to these tests. From what I recall, when a namespace is changed by the operator, it will restart all of the supported resources (with the exception of paused deployments). It does not check if the resource is also annotated.

The use cases seem good. One suggestion is to have multiple of each resource so we can verify that the operator isn't just updating all deployments and doesn't modify resources not in its configuration.

integration-tests/manifests/sample-deployment.yaml
lisguo: agreed, they should all use the same sample "app". Which it seems like in your case I would just use nginx

integration-tests/manifests/sample-statefulset.yaml
lisguo: yeah just 1 operator for the entire cluster -- and workloads in a sample namespace.

zhihonl:
I don't see a link to a sample test run. Can you add that in the description.

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
+package annotations
```
zhihonl, line 1: For future reference, I suggest breaking code like this into separate PRs especially when we know they are separated test cases.

In the description we have 4 main use cases then I'm assuming this PR can be broken into 4 PRs. 1000 lines of code review is difficult to track of.

zhihonl: nit: Idk if all the tests are dependent on each other. but if not it would be better to divide into `Deployment`, `Daemonset`,`Namespace`,and `Statefulset`, so something like `validate_annotations_deployment_test.go`,`validate_annotations_daemonset_test.go`, etc

This way when making a test change for a specific k8s component, it's more clear instead of having to scroll through other unrelated tests.

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
...
+const amazonCloudwatchNamespace = "amazon-cloudwatch"
+
+const daemonSetName = "sample-daemonset"
+
+const amazonControllerManager = "cloudwatch-controller-manager"
+
+var opMutex sync.Mutex
+
+// ---------------------------USE CASE 1 (Java and Python on Deployment) ----------------------------------------------
+func TestUseCase1(t *testing.T) {
```
zhihonl, line 40: nit: Better function naming such as `TestJavaAndPythonDeployment`. This helps with `go test` output instead of having to go back and read the comment and remember what the test was for

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
...
+
+	t.Parallel()
+	clientSet := setupTest(t)
+	uniqueNamespace := "sample-namespace-1"
```
zhihonl, line 44: nit: namespace naming should be close to what the test is accomplishing. Ex: `deployment-namespace-java-python` for test case 1, `deployment-namespace-java-only` for test case 2, etc.

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
...
+	deploymentPods, err := clientSet.CoreV1().Pods(uniqueNamespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: set.AsSelector().String(),
+	})
+	if err != nil {
+		t.Errorf("Error listing pods for deployment: %s", err.Error())
+	}
+
+	//wait for pods to update
+	if !checkIfAnnotationExists(deploymentPods, []string{injectJavaAnnotation, autoAnnotateJavaAnnotation, injectPythonAnnotation, autoAnnotatePythonAnnotation}) {
+		t.Error("Incorrect Annotations")
```
zhihonl, line 102: nit: Better error message is "Missing Java and Python annotations". Same for all other error messages for other test cases as well

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
...
+		}
+	}
+	fmt.Println("Annotations are correct!")
+	return true
+}
+func updateAnnotationConfig(deployment *appsV1.Deployment, jsonStr string) {
+	args := deployment.Spec.Template.Spec.Containers[0].Args
+	indexOfAutoAnnotationConfigString := findIndexOfPrefix("--auto-annotation-config=", args)
+	//if auto annotation not part of config, we will add it
+	if indexOfAutoAnnotationConfigString < 0 || indexOfAutoAnnotationConfigString >= len(deployment.Spec.Template.Spec.Containers[0].Args) {
```
zhihonl, line 969: What is the point of checking `indexOfAutoAnnotationConfigString >= len(deployment.Spec.Template.Spec.Containers[0].Args` here? If  Doesn't `findIndexOfPrefix` function only return -1 when the index is not found?

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
...
+		}
+	}
+	fmt.Println("Namespace annotations are correct!")
+	return true
+}
+func updateOperator(t *testing.T, clientSet *kubernetes.Clientset, deployment *appsV1.Deployment) bool {
+	opMutex.Lock() // Lock the mutex at the beginning
+	defer opMutex.Unlock()
+	var err error
+	args := deployment.Spec.Template.Spec.Containers[0].Args
```
zhihonl, line 933: What's the reason that we can't just update operator using the `deployment` variable as it is if it's already updated with the new arg?

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
...
+		}
+	}
+
+	// List pods belonging to the deployment
+	set := labels.Set(deployment.Spec.Selector.MatchLabels)
+	deploymentPods, err := clientSet.CoreV1().Pods(uniqueNamespace).List(context.TODO(), metav1.ListOptions{
+		LabelSelector: set.AsSelector().String(),
+	})
+	if err != nil {
+		t.Error("Failed to update Operator")
```
zhihonl, line 167: Can we also append the error message here in case the API call is broken and we know what was the issue.

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
...
+	//finding where index of --auto-annotation-config= is (if it doesn't exist it will be appended)
+	updateAnnotationConfig(deployment, string(jsonStr))
+	if !updateOperator(t, clientSet, deployment) {
+		t.Error("Failed to update Operator")
+	}
+
+	//check if deployment has annotations.
+	deployment, err = clientSet.AppsV1().Deployments(uniqueNamespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
+	if err != nil {
+		if err != nil {
```
zhihonl, line 156: Why do we have double if statements with the same conditions?

zhihonl:
Did we add the tests coverage that @sky333999  asked for or is that being done in a separate PR?

> I think the use cases we have are a good start, but I think we are still missing a few to give us confidence. We can get a framework ready for the initial commit and then incrementally add more use cases.
> 
> Couple of the use cases that I see are missing:
> 
> * If a namespace is annotated, confirm that only apps that are not already annotated get restarted and now have annotations. i.e. Existing annotated apps should NOT be restarted.
> * If a resource is already annotated, what is the behavior of adding it in the operator args as well?
> * If a resource is manually annotated and now its namespace is added for auto annotation, the resource should not be modified and should not be restarted.
> * If a resource that is auto annotated is manually patched to remove the annotations, our mutator should add back the annotations.

integration-tests/manifests/annotations/validate_annotations_test.go
```diff
@@ -0,0 +1,1003 @@
...
+	}
+
+	//finding where index of --auto-annotation-config= is (if it doesn't exist it will be appended)
+	updateAnnotationConfig(deployment, string(jsonStr))
+	if !updateOperator(t, clientSet, deployment) {
+		t.Error("Failed to update Operator")
+	}
+
+	//check if deployment has annotations.
+	deployment, err = clientSet.AppsV1().Deployments(uniqueNamespace).Get(context.TODO(), deploymentName, metav1.GetOptions{})
```
zhihonl, line 154: Everything from getting deployment to verifying if the annotation exists are repeated code. I suggest making a helper function that just combine these calls for you then you can reduce the code size by a lot.

integration-tests/manifests/annotations/validate_annotations_test.go
Paramadon: That's correct! I'll change that.

Paramadon:
> Did we add the tests coverage that @sky333999 asked for or is that being done in a separate PR?
> 
> > I think the use cases we have are a good start, but I think we are still missing a few to give us confidence. We can get a framework ready for the initial commit and then incrementally add more use cases.
> > Couple of the use cases that I see are missing:
> > 
> > * If a namespace is annotated, confirm that only apps that are not already annotated get restarted and now have annotations. i.e. Existing annotated apps should NOT be restarted.
> > * If a resource is already annotated, what is the behavior of adding it in the operator args as well?
> > * If a resource is manually annotated and now its namespace is added for auto annotation, the resource should not be modified and should not be restarted.
> > * If a resource that is auto annotated is manually patched to remove the annotations, our mutator should add back the annotations.

It will be done in a seperate pr since this has the base cases we can do another pr to get any other cases

.github/workflows/operator-integration-test.yml
```diff
@@ -110,3 +110,15 @@ jobs:
...
           kubectl delete instrumentation sample-instrumentation
+
+      - name: Test Annotations
+        run: |
+          kubectl apply -f integration-tests/manifests/sample-deployment.yaml
+          kubectl apply -f integration-tests/manifests/sample-daemonset.yaml
+          kubectl apply -f integration-tests/manifests/sample-statefulset.yaml
+          kubectl get pods -A
+          kubectl describe pods -n default
+          go test ./integration-tests/manifests/annotations -timeout 30m
```
zhihonl, line 121: Can you add `go test -v` so it's verbose mode that way we can see each test output

Paramadon: will do!

zhihonl (approved):

lisguo (approved):

PR #143, opened 2024-04-03
Always insert mutations for resources in scope and restart only required resources in namespace scope

sky333999:
*Description of changes:* 
1) If a resource is in scope for auto-annotation per the args passed to the operator and the resource itself on the cluster is in a state such that it has only the auto-annotate annotation or the instr-inject annotation, we would previously skip such resources. Updating this logic to add the missing annotation in such scenarios. Post these changes, the behavior for auto-annotation can be summarized as:
* If a resource is in the operator args and does not already have either auto-instr or auto-annotate annotations, add them
* If a resource is not in the operator args and has both annotations, remove both the annotations
* If a resource is not in the operator args and has only one annotation, it implies it was manually annotated, so leave it untouched

2) When a namespace is auto-annotated, the current logic restarts all the resources in the namespace. Optimizing this logic to only restart resources that do NOT already have the auto-instr annotation.
This prevents an unnecessary restart for:
* Resources in the namespace that were manually tagged with auto-instr annotation
* Resources in the namespace that were explicitly passed in to the operator args and hence was already restarted when the operator added the annotations

### DIFF
--- a/pkg/instrumentation/annotationmutator.go
+++ b/pkg/instrumentation/annotationmutator.go
@@ -18,9 +18,6 @@ type insertAnnotationMutation struct {
 }
 
 func (m *insertAnnotationMutation) Mutate(annotations map[string]string) bool {
-	if !m.shouldMutate(annotations) {
-		return false
-	}
 	var mutated bool
 	for key, value := range m.insert {
 		if _, ok := annotations[key]; !ok {
@@ -29,15 +26,6 @@ func (m *insertAnnotationMutation) Mutate(annotations map[string]string) bool {
 		}
 	}
 	return mutated
-}
-
-func (m *insertAnnotationMutation) shouldMutate(annotations map[string]string) bool {
-	for key := range m.insert {
-		if _, ok := annotations[key]; ok {
-			return false
-		}
-	}
-	return true
 }
 
 // NewInsertAnnotationMutation creates a new mutation that inserts annotations. All provided annotation keys
--- a/pkg/instrumentation/annotationmutator.go
+++ b/pkg/instrumentation/annotationmutator.go
@@ -9,27 +9,30 @@ import (
 
 // AnnotationMutation is used to modify an annotation map.
 type AnnotationMutation interface {
-	// Mutate attempts to modify the annotations map. Returns whether the function changed the annotations.
-	Mutate(annotations map[string]string) bool
+	// Mutate attempts to modify the annotations map. Returns the mutated annotations if any and whether the function
+	// changed the annotations.
+	Mutate(annotations map[string]string) (map[string]string, bool)
 }
 
 type insertAnnotationMutation struct {
 	insert map[string]string
 }
 
-func (m *insertAnnotationMutation) Mutate(annotations map[string]string) bool {
+func (m *insertAnnotationMutation) Mutate(annotations map[string]string) (map[string]string, bool) {
 	var mutated bool
+	mutatedAnnotations := make(map[string]string)
 	for key, value := range m.insert {
 		if _, ok := annotations[key]; !ok {
 			annotations[key] = value
+			mutatedAnnotations[key] = value
 			mutated = true
 		}
 	}
-	return mutated
+	return mutatedAnnotations, mutated
 }
 
-// NewInsertAnnotationMutation creates a new mutation that inserts annotations. All provided annotation keys
-// must be present for it to attempt to insert.
+// NewInsertAnnotationMutation creates a new mutation that inserts annotations. Any missing annotation key
+// is sufficient for it to attempt to insert.
 func NewInsertAnnotationMutation(annotations map[string]string) AnnotationMutation {
 	return &insertAnnotationMutation{insert: annotations}
 }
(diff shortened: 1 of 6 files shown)


pkg/instrumentation/auto/annotation.go
```diff
@@ -37,25 +37,27 @@ type AnnotationMutators struct {
...
 // MutateAndPatchAll runs the mutators for each of the supported resources and patches them.
 func (m *AnnotationMutators) MutateAndPatchAll(ctx context.Context) {
 	mutateAndPatchFunc := m.patchFunc(ctx, m.MutateObject)
-	m.rangeObjectList(ctx, &corev1.NamespaceList{}, &client.ListOptions{},
-		chainCallbacks(mutateAndPatchFunc, m.restartNamespaceFunc(ctx)),
-	)
 	m.rangeObjectList(ctx, &appsv1.DeploymentList{}, &client.ListOptions{}, mutateAndPatchFunc)
 	m.rangeObjectList(ctx, &appsv1.DaemonSetList{}, &client.ListOptions{}, mutateAndPatchFunc)
 	m.rangeObjectList(ctx, &appsv1.StatefulSetList{}, &client.ListOptions{}, mutateAndPatchFunc)
+	m.rangeObjectList(ctx, &corev1.NamespaceList{}, &client.ListOptions{},
```
sky333999, line 58: Flipped the order to mutate the resources first (triggering a restart) and then mutate the namespace such that this time, a 2nd restart can be avoided.

pkg/instrumentation/auto/callback.go
```diff
@@ -95,3 +101,42 @@ func (m *AnnotationMutators) restartNamespaceFunc(ctx context.Context) objectCal
...
+			return m.checkDifferentInjectAnnotations(namespace.GetObjectMeta(), o.Spec.Template.GetObjectMeta())
+		default:
+			return false
+		}
+	}
+}
+
+// checkDifferentInjectAnnotations returns true if both the objects do NOT have identical inject annotations and false
+// otherwise. If both objects do not have any inject annotations, return false.
+func (m *AnnotationMutators) checkDifferentInjectAnnotations(obj1, obj2 metav1.Object) bool {
```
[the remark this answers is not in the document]
sky333999: Yeah thats a good point. Also realized the way I had it wouldnt handle yet another edge case for removal of annotations. So I restructured the logic to pass along the mutated annotations (inserts & removals) and use those to determine whether or not we need a restart.

pkg/instrumentation/annotationmutator.go
```diff
@@ -9,39 +9,30 @@ import (
...
+	mutatedAnnotations := make(map[string]string)
 	for key, value := range m.insert {
 		if _, ok := annotations[key]; !ok {
 			annotations[key] = value
+			mutatedAnnotations[key] = value
 			mutated = true
 		}
 	}
-	return mutated
+	return mutatedAnnotations, mutated
```
jefchien, line 31: nit: Could remove the mutated bool or could change it to not return a bool at all since the presence of entries in the map would indicate a change

```suggestion
	return mutatedAnnotations, len(mutatedAnnotations) != 0
```

pkg/instrumentation/annotationmutator.go
```diff
@@ -88,18 +81,24 @@ func NewAnnotationMutator(mutations []AnnotationMutation) AnnotationMutator {
...
 	if annotations == nil {
 		annotations = make(map[string]string)
 	}
 	var anyMutated bool
+	allMutatedAnnotations := make(map[string]string)
 	for _, mutation := range m.mutations {
-		mutated := mutation.Mutate(annotations)
+		mutatedAnnotations, mutated := mutation.Mutate(annotations)
 		anyMutated = anyMutated || mutated
+		if mutated {
```
jefchien, line 96: nit: Don't need the check. An empty or even nil map will just not do anything with a for-range loop.

pkg/instrumentation/auto/annotation.go
```diff
@@ -37,29 +37,38 @@ type AnnotationMutators struct {
 	daemonSetMutators   map[string]instrumentation.AnnotationMutator
 	statefulSetMutators map[string]instrumentation.AnnotationMutator
 	defaultMutator      instrumentation.AnnotationMutator
+	injectAnnotations   map[string]struct{}
 }
 
 // RestartNamespace sets the restartedAtAnnotation for each of the namespace's supported resources and patches them.
-func (m *AnnotationMutators) RestartNamespace(ctx context.Context, namespace *corev1.Namespace) {
+func (m *AnnotationMutators) RestartNamespace(ctx context.Context, namespace *corev1.Namespace, mutatedAnnotations map[string]string) {
 	restartAndPatchFunc := m.patchFunc(ctx, setRestartAnnotation)
```
jefchien, line 45: nit: Move chain callback declaration to here
```suggestion
	restartAndPatchFunc := chainCallbacks(m.shouldRestartFunc(mutatedAnnotations), m.patchFunc(ctx, setRestartAnnotation))
```

pkg/instrumentation/auto/callback.go
```diff
@@ -60,38 +67,88 @@ func createPatch(obj client.Object) (client.Patch, error) {
...
+			return ret, false
 		}
-		return true
+		return ret, true
 	}
 }
 
 func (m *AnnotationMutators) restartNamespaceFunc(ctx context.Context) objectCallbackFunc {
-	return func(obj client.Object) bool {
+	return func(obj client.Object, mutatedAnnotations_ any) (any, bool) {
```
jefchien, line 97: super minor nit: Consider renaming field. If it really can be anything, then at the moment of calling this function, you won't know if it's `mutatedAnnotations` or whatever. Maybe `previousResult` or `passToNext` like you have above. I would just try to keep it consistent.

jefchien (dismissed):

lisguo (dismissed):

pkg/instrumentation/auto/callback.go
```diff
@@ -60,38 +67,88 @@ func createPatch(obj client.Object) (client.Patch, error) {
...
+	return func(obj client.Object, _ any) (any, bool) {
 		patch, err := createPatch(obj)
 		if err != nil {
 			m.logger.Error(err, "Unable to create patch",
 				"kind", fmt.Sprintf("%T", obj),
 				"name", obj.GetName(),
 				"namespace", obj.GetNamespace(),
 			)
-			return false
+			return nil, false
```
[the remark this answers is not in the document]
sky333999: Right. We had to pass quite a few nils due to the signature change for the callback. But I did test the changes on an EKS cluster and do not see any errors.

lisguo (approved):

jefchien (approved):

PR #154, opened 2024-04-17
Implement tests for annotation validations

Paramadon:
*Issue #, if available:*
The Enablement experience integ tests would fail sometimes due to a race condition.

*Description of changes:*
Removed concurrency and instead seperated tests into seperate minikubes running concurrently and avoiding any weird race conditions to occur(also decreases time from 17 minutes to only 2-3 minutes for the annotation test). 

### Below is a picture of one test running locally 100 times and all test pass

Screenshot 2024-04-18 at 2 17 43 PM

### Below is a picture of the full integ test running 5 times and it passed all times
Screenshot 2024-04-17 at 5 54 56 PM

## Deployments Cases
Use Case 1: Java and Python on Deployment - Both Python and Java should be annotated
Use Case 2: Java on Deployment and Python Should be Removed since both are injected on the deployment
Use Case 3: Python on Deployment and Java annotations should be removed from deployment

## Daemonset Cases
Use Case 4: Python and Java on DaemonSet - Both Python and Java should be on the Daemonset
Use Case 5: Java on DaemonSet and Python should be removed
Use Case 6: Python on DaemonSet Java annotation should be removed

## Namespace Cases
Use Case 7: Both Java and Python on Namespace - All pods in the name space should have both Python and Java annotations
Use Case 8: Java on Namespace and Python should be removed 
Use Case 9: Python on Namespace and Java annotation should be removed

## Statefulset Cases

Use Case 10: Python and Java on Stateful set 
Use Case 11: Java on Stateful set and Python should be removed
Use Case 12: Python on Stateful set and java should be removed


.github/workflows/operator-integration-test.yml
```diff
@@ -110,3 +110,202 @@ jobs:
           cd integration-tests/manifests/cmd
           go run validate_instrumentation_vars.go default ns_instrumentation_env_variables.json
           kubectl delete instrumentation sample-instrumentation
+
+
+
+  DeploymentAnnotationsTest:
```
lisguo, line 116: non blocking nit: now that we have these annotations test, we can change the original integration test job to be called "InstrumentationTest" or something.

.github/workflows/operator-integration-test.yml
```diff
@@ -110,3 +110,202 @@ jobs:
...
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Start minikube
```
lisguo, line 178: nit: you can put steps "Start minikube" to "Deploy operator to minikube" as a reusable workflow and call it in all the tests. https://www.dhiwise.com/post/the-ultimate-guide-to-github-reusable-workflows-maximize-efficiency-and-collaboration

integration-tests/manifests/annotations/validate_annotation_methods.go
```diff
@@ -0,0 +1,370 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package annotations
```
lisguo, line 4: super nit: I would rename the files to be shorter since it's self explanatory they are related to annotations since it's in the annotations package. Something like just `validate_daemonset.go`.

also maybe call this one specifically `validate_common.go`

integration-tests/util/util.go
```diff
@@ -0,0 +1,74 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package util
```
lisguo, line 4: nice!

lisguo (approved):

lisguo:
Also I would rename the PR to "Implement tests for annotation validations" or something similar. We reverted the commit so it's not really a "fix" anymore but a reimplementation

zhihonl (approved):

PR #161, opened 2024-04-26
Adding Extra Use Cases for Enablement Experience Integration test 

Paramadon:
*Issue #, if available:*

The Enablement experience integration test needed some extra use cases to make sure proper auto annotations behavior is occuring. So in order to satisfy this, more test cases were added to ensure correct behavior.

*Description of changes:*

### TestAnnotationsOnMultipleResources
- If there are multiple resources on the same namespace, all of non-annotated resources should be annotated.

### TestAutoAnnotationForManualAnnotationRemoval
-  If a resource that is auto annotated is manually patched to remove the annotations, our mutator should add back the annotations.

### TestOnlyNonAnnotatedAppsShouldBeRestarted
-  If a resource is manually annotated and now its namespace is added for auto annotation, the resource should not be modified and should not be restarted.
- If a namespace is annotated, confirm that only apps that are not already annotated get restarted and now have annotations. Existing annotated apps should NOT be restarted.

### TestAlreadyAutoAnnotatedResourceShouldNotRestart

- If a resource is already annotated and then operator adds annotation, auto-annotation annotation is not added

### DIFF
--- a/integration-tests/manifests/annotations/validate_annotation_methods.go
+++ b/integration-tests/manifests/annotations/validate_annotation_methods.go
@@ -37,7 +37,7 @@ const (
 
 	daemonSetName = "sample-daemonset"
 
-	amazonControllerManager = "amazon-cloudwatch-observability-controller-manager"
+	amazonControllerManager = "cloudwatch-controller-manager"
 
 	sampleDaemonsetYamlRelPath      = "../sample-daemonset.yaml"
 	sampleDeploymentYamlNameRelPath = "../sample-deployment.yaml"
@@ -390,12 +390,10 @@ func setupFunction(t *testing.T, namespace string, apps []string) (*kubernetes.C
 	if err := createNamespaceAndApplyResources(t, clientSet, uniqueNamespace, apps); err != nil {
 		t.Fatalf("Failed to create/apply resoures on namespace: %v", err)
 	}
-	fmt.Println("setupfunction", namespace, apps)
 	t.Cleanup(func() {
 		if err := deleteNamespaceAndResources(clientSet, uniqueNamespace, apps); err != nil {
 			t.Fatalf("Failed to delete namespaces/resources: %v", err)
 		}
-		fmt.Println("HIIII")
 	})
 
 	return clientSet, uniqueNamespace
--- a/integration-tests/manifests/annotations/validate_annotation_methods.go
+++ b/integration-tests/manifests/annotations/validate_annotation_methods.go
@@ -6,6 +6,8 @@ package annotations
 import (
 	"context"
 	"fmt"
+	"github.com/google/uuid"
+	"strconv"
 	"testing"
 
 	"github.com/aws/amazon-cloudwatch-agent-operator/integration-tests/util"
@@ -35,7 +37,7 @@ const (
 
 	daemonSetName = "sample-daemonset"
 
-	amazonControllerManager = "cloudwatch-controller-manager"
+	amazonControllerManager = "amazon-cloudwatch-observability-controller-manager"
 
 	sampleDaemonsetYamlRelPath      = "../sample-daemonset.yaml"
 	sampleDeploymentYamlNameRelPath = "../sample-deployment.yaml"
@@ -105,7 +107,6 @@ func createNamespace(clientSet *kubernetes.Clientset, name string) error {
 	if err == nil {
 		return nil
 	} else if !errors.IsNotFound(err) {
-		fmt.Println("Some other error")
 		return err
 	}
 
@@ -256,8 +257,6 @@ func checkIfAnnotationExists(clientset *kubernetes.Clientset, pods *v1.PodList, 
 		}
 
 		fmt.Println("Annotations not found in all pods or some pods are not in Running phase. Retrying...")
-		fmt.Println("We are in 4")
-
 		time.Sleep(timeBetweenRetries)
 	}
 }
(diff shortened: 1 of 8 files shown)


musa-asad (approved):

integration-tests/manifests/annotations/validate_annotations_namespace_test.go
```diff
@@ -127,3 +132,300 @@ func TestPythonOnlyNamespace(t *testing.T) {
...
+	uniqueNamespace := fmt.Sprintf("multiple-resources-%d", randomNumber)
+	if err := createNamespaceAndApplyResources(t, clientSet, uniqueNamespace, []string{sampleDeploymentYamlNameRelPath}); err != nil {
+		t.Fatalf("Failed to create/apply resoures on namespace: %v", err)
+	}
+	defer func() {
+		if err := deleteNamespaceAndResources(clientSet, uniqueNamespace, []string{sampleDeploymentYamlNameRelPath}); err != nil {
+			t.Fatalf("Failed to delete namespaces/resources: %v", err)
+		}
+	}()
+	if err := createNamespaceAndApplyResources(t, clientSet, uniqueNamespace, []string{sampleNginxAppYamlNameRelPath}); err != nil {
```
jefchien, line 280: Could this be created as part of the first `createNamespaceAndApplyResources` in the slice parameter?

integration-tests/manifests/annotations/validate_annotations_namespace_test.go
```diff
@@ -127,3 +132,300 @@ func TestPythonOnlyNamespace(t *testing.T) {
...
+	uniqueNamespace := fmt.Sprintf("multiple-resources-%d", randomNumber)
+	startTime := time.Now()
+	if err := createNamespaceAndApplyResources(t, clientSet, uniqueNamespace, []string{sampleDeploymentYamlNameRelPath}); err != nil {
+		t.Fatalf("Failed to create/apply resoures on namespace: %v", err)
+	}
+	defer func() {
+		if err := deleteNamespaceAndResources(clientSet, uniqueNamespace, []string{sampleDeploymentYamlNameRelPath}); err != nil {
+			t.Fatalf("Failed to delete namespaces/resources: %v", err)
+		}
+	}()
```
jefchien, line 364: This seems like a common block of setup code. Consider moving it into a helper function.

integration-tests/manifests/annotations/validate_annotation_methods.go
```diff
@@ -368,3 +379,29 @@ func checkResourceAnnotations(t *testing.T, clientSet *kubernetes.Clientset, res
...
+	// Check if the revision history contains revisions after the first revision
+	lines := strings.Split(string(output), "\n")
+	fmt.Println("Below is the lines and its count")
+	fmt.Println(lines)
+	fmt.Println(len(lines))
+	for _, line := range lines {
+		fmt.Println(line)
+	}
+	//4 lines are the text above the revisions
+	return len(lines) - 4, nil
```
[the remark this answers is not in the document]
Paramadon: There isn't a direct way to get the number of revisions and I tried to do it using timestamp but that based off the creation time so it doesn't account for restart. But instead I'll change this to use kubectl kubectl rollout history deployment sample-deployment --revision int and see which number fails which will give me the exact number of revisions. I think this is a better so I will change it

integration-tests/manifests/annotations/validate_annotations_namespace_test.go
Paramadon: yes, I will make them into one function

integration-tests/manifests/annotations/validate_annotations_namespace_test.go
Paramadon: makes sense, I put everything above the defer func in the helper. I changed the defer func to t.cleanup and I have to make sure it is in the original test function otherwise the deployment will be deleted, but the rest is in helper func.

integration-tests/manifests/annotations/validate_annotation_methods.go
```diff
@@ -56,8 +61,9 @@ func createNamespaceAndApplyResources(t *testing.T, clientset *kubernetes.Client
 	}
 
 	for _, file := range resourceFiles {
-		err = applyYAMLWithKubectl(filepath.Join("..", file), name)
+		err = applyYAMLWithKubectl(file, name)
 		if err != nil {
+			fmt.Println(name, file)
 			t.Error("Could not apply resources")
```
jefchien, line 67: nit: Could do
```suggestion
			t.Errorf("Could not apply resources %s/%s", name, file)
```

integration-tests/manifests/annotations/validate_annotations_namespace_test.go
```diff
@@ -127,3 +133,221 @@ func TestPythonOnlyNamespace(t *testing.T) {
 		t.Error("Missing Python annotations")
 	}
 }
+
+// Multiple resources on the same namespace should all get annotations
+func TestAnnotationsOnMultipleResources(t *testing.T) {
+
+	clientSet := setupTest(t)
+	newUUID := uuid.New()
+	uniqueNamespace := fmt.Sprintf("multiple-resources-%v", fmt.Sprint(newUUID))
```
jefchien, line 142: nit:
```suggestion
	uniqueNamespace := fmt.Sprintf("multiple-resources-%s", uuid.NewString())
```

integration-tests/manifests/annotations/validate_annotation_methods.go
```diff
@@ -368,3 +378,35 @@ func checkResourceAnnotations(t *testing.T, clientSet *kubernetes.Clientset, res
...
+}
+
+func setupFunction(t *testing.T, namespace string, apps []string) (*kubernetes.Clientset, string) {
+	clientSet := setupTest(t)
+	newUUID := uuid.New()
+	uniqueNamespace := fmt.Sprintf(namespace+"-%v", fmt.Sprint(newUUID))
+	if err := createNamespaceAndApplyResources(t, clientSet, uniqueNamespace, apps); err != nil {
+		t.Fatalf("Failed to create/apply resoures on namespace: %v", err)
+	}
+
```
jefchien, line 393: Could you add the `t.Cleanup` here?

musa-asad (approved):

jefchien (approved):

PR #162, opened 2024-04-26
Modify operator integration test to use yaml instead of helm templating.

musa-asad:
*Description of changes:*
Replaced the helm templating to use yaml instead.

### DIFF
--- a/integration-tests/manifests/cloudwatch-agent-daemonset.yaml
+++ b/integration-tests/manifests/cloudwatch-agent-daemonset.yaml
@@ -6,7 +6,7 @@ metadata:
 spec:
   mode: daemonset
   serviceAccount: cloudwatch-agent
-  image: public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.300031.1b317
+  image: public.ecr.aws/cloudwatch-agent/cloudwatch-agent:latest
   config: |
     {
       "agent": {
--- a/integration-tests/manifests/cloudwatch-agent-daemonset.yaml
+++ b/integration-tests/manifests/cloudwatch-agent-daemonset.yaml
@@ -1,0 +1,80 @@
+apiVersion: cloudwatch.aws.amazon.com/v1alpha1
+kind: AmazonCloudWatchAgent
+metadata:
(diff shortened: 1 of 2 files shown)


sethAmazon (approved):

sethAmazon (dismissed):

integration-tests/manifests/cloudwatch-agent-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
+apiVersion: cloudwatch.aws.amazon.com/v1alpha1
```
lisguo, line 1: is there a reason we are not using the template in the helm chart? https://github.com/aws-observability/helm-charts/blob/main/charts/amazon-cloudwatch-observability/templates/linux/cloudwatch-agent-daemonset.yaml

integration-tests/manifests/cloudwatch-agent-daemonset.yaml
```diff
@@ -0,0 +1,80 @@
...
+kind: AmazonCloudWatchAgent
+metadata:
+  name: cloudwatch-agent
+  namespace: amazon-cloudwatch
+spec:
+  mode: daemonset
+  serviceAccount: cloudwatch-agent
+  image: public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.300031.1b317
```
lisguo, line 9: should we use the latest cwagent? might want to ensure we are compatible with the latest released agent

`public.ecr.aws/cloudwatch-agent/cloudwatch-agent:latest`

integration-tests/manifests/cloudwatch-agent-daemonset.yaml
musa-asad: I think we can use that template, but it would require passing in a lot of values, for which it seemed more frugal to just have set up already.

integration-tests/manifests/cloudwatch-agent-daemonset.yaml
musa-asad: I will fix.

sethAmazon (approved):

PR #163, opened 2024-04-30
[DO NOT MERGE] Update app_signals to application_signals

bjrara:
*Issue #, if available:*
Rename `app_signals` to `application_signals` for GA release. It's part of the PR: https://github.com/aws/amazon-cloudwatch-agent/pull/1133.

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,11 +136,11 @@ spec:
         value: "none"
       - name: OTEL_LOGS_EXPORTER
         value: "none"
-      - name: OTEL_AWS_APP_SIGNALS_ENABLED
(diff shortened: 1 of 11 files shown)


lisguo:
The change to be made to https://github.com/aws-observability/helm-charts/.

I think also we need to ensure that we pluck both app_signals and application_signals if they exist for backwards compatibility?

bjrara:
> The change to be made to https://github.com/aws-observability/helm-charts/.
> 
> I think also we need to ensure that we pluck both app_signals and application_signals if they exist for backwards compatibility?

Thanks! Will do.

PR #165, opened 2024-05-02
[Enablement Experience] Avoiding failure edge case

Paramadon:
*Issue #, if available:*
There is a small fail case that can happen during this test which is unlikely maybe 10% of the time. The test would fail because the operator would change the deployment while midst updating deployment.
*Description of changes:*
This change just adds a wait condition for the deployment to be fully ready

### DIFF
--- a/integration-tests/manifests/annotations/validate_annotations_namespace_test.go
+++ b/integration-tests/manifests/annotations/validate_annotations_namespace_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent-operator/integration-tests/util"
 	"github.com/aws/amazon-cloudwatch-agent-operator/pkg/instrumentation/auto"
 	"github.com/google/uuid"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"math/big"
 	"os"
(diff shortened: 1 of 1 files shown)


integration-tests/manifests/annotations/validate_annotations_namespace_test.go
```diff
@@ -186,22 +186,38 @@ func TestAutoAnnotationForManualAnnotationRemoval(t *testing.T) {
...
+
+		if deployment.Status.AvailableReplicas == *deployment.Spec.Replicas && deployment.Status.UpdatedReplicas == *deployment.Spec.Replicas {
+			if deployment.Status.Replicas == deployment.Status.AvailableReplicas {
+				fmt.Println("All pods are fully ready and no pods are terminating.")
+				break
+			}
+		}
+
+		// Sleep for a short interval before checking again
+		time.Sleep(5 * time.Second)
```
sethAmazon, line 215: Should we be doing wait groups instead of sleep?

Paramadon: Hmm I think wait groups might not be the best approach for this scenario because we're dealing with a loop that waits for a certain condition to be met, rather than coordinating multiple concurrent tasks.

Paramadon: Wait groups might not be the most suitable approach for this scenario because we're dealing with a loop that waits for a certain condition to be met, rather than coordinating multiple concurrent tasks. It is a single goroutine that runs in a loop.

PR #171, opened 2024-05-09
Adding GPU E2E Test

Paramadon:
*Issue #, if available:*
Currently there is an integration test for GPU, but we need an e2e test for this.
*Description of changes:*
The main difference between integration test and this is that we install the eks addon instead of mocking the dcgm exporter and daemonsets/services. 
### Steps happening in the e2e
1. Creating beta cluster with node groups that has gpu instance and gpu ami
2. installing eks addon
3. Running gpu test

### E2E Test
https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/9033457562


musa-asad (dismissed):

integration-tests/terraform/gpu/main.tf
```diff
@@ -0,0 +1,146 @@
...
+      aws_eks_node_group.this,
+      aws_eks_addon.this
+  ]
+
+  provisioner "local-exec" {
+    command = <<EOT
+      go test ${var.test_dir} -eksClusterName ${aws_eks_cluster.this.name} -computeType=EKS -v -eksDeploymentStrategy=DAEMON -eksGpuType=nvidia
+
+      # Get all pods and describe them
+      kubectl get pods --all-namespaces -o wide > pods.txt
```
movence, line 137: why do you need to write it to a file?

integration-tests/gpu/gpu_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/stretchr/testify/suite"
+
+	"github.com/aws/amazon-cloudwatch-agent-test/environment"
+	"github.com/aws/amazon-cloudwatch-agent-test/environment/computetype"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/metric/dimension"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/status"
+	"github.com/aws/amazon-cloudwatch-agent-test/test/test_runner"
+)
+
+type GPUTestSuite struct {
```
movence, line 22: As we discussed, you don't need a test suite for this e2e test. The purpose of test suite is to group similar tests into a group to execute them all in the suite. There is no other test will be added to this suite. Other accelerated compute instance types (eg tranium) will be its own test with a different instance type in terraform file that will execute a different test.

PR #184, opened 2024-06-07
Add JMX Annotations

sethAmazon:
*Issue #, if available:*

*Description of changes:*


dchappa (dismissed):

sethAmazon:
The ticket did not mention Kafka metric. They are needed. Added them to this pr thus required resetting the approvals.

jefchien:
Change the destination branch to `main-jmx`. Don't want to put it into main yet.

pkg/instrumentation/defaultinstrumentation.go
```diff
@@ -43,6 +43,26 @@ func getDefaultInstrumentation(agentConfig *adapters.CwaConfig) (*v1alpha1.Instr
...
+	javaEnvVar := []corev1.EnvVar{
+		{Name: "OTEL_AWS_APP_SIGNALS_ENABLED", Value: "true"}, //TODO: remove in favor of new name once safe
+		{Name: "OTEL_AWS_APPLICATION_SIGNALS_ENABLED", Value: "true"},
+		{Name: "OTEL_TRACES_SAMPLER_ARG", Value: "endpoint=http://cloudwatch-agent.amazon-cloudwatch:2000"},
+		{Name: "OTEL_TRACES_SAMPLER", Value: "xray"},
+		{Name: "OTEL_EXPORTER_OTLP_PROTOCOL", Value: "http/protobuf"},
+		{Name: "OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", Value: exporterPrefix + "cloudwatch-agent.amazon-cloudwatch:4316/v1/traces"},
+		{Name: "OTEL_AWS_APP_SIGNALS_EXPORTER_ENDPOINT", Value: exporterPrefix + "cloudwatch-agent.amazon-cloudwatch:4316/v1/metrics"}, //TODO: remove in favor of new name once safe
+		{Name: "OTEL_AWS_APPLICATION_SIGNALS_EXPORTER_ENDPOINT", Value: exporterPrefix + "cloudwatch-agent.amazon-cloudwatch:4316/v1/metrics"},
+		{Name: "OTEL_METRICS_EXPORTER", Value: "none"},
```
jefchien, line 55: This cannot be "none" for the JMX case. Setting it to "none" will disable all metrics from being exported. We only need to set it to "none" if JMX isn't enabled for the pod.

pkg/instrumentation/podmutator.go
```diff
@@ -402,14 +404,42 @@ func (pm *instPodMutator) selectInstrumentationInstanceFromNamespace(ctx context
...
+	targetSystems = getTargetSystems(ns, pod, annotationJMXKafkaProducer, targetSystems)
+	if targetSystems != "" {
+		return map[string]string{
+			"OTEL_EXPORTER_OTLP_METRICS_ENDPOINT":            "http://cloudwatch-agent.amazon-cloudwatch:4314/v1/metrics",
+			"OTEL_INSTRUMENTATION_RUNTIME_TELEMETRY_ENABLED": "false",
+			"OTEL_JMX_ENABLED":                               "true",
+			"OTEL_JMX_TARGET_SYSTEM":                         targetSystems,
+		}
+	}
+	return make(map[string]string)
```
jefchien, line 429: No point in creating an empty map.
```suggestion
	return nil
```

pkg/instrumentation/podmutator.go
```diff
@@ -402,14 +404,42 @@ func (pm *instPodMutator) selectInstrumentationInstanceFromNamespace(ctx context
...
+}
+
+func getTargetSystems(ns corev1.Namespace, pod corev1.Pod, annotation string, targets string) string {
```
jefchien, line 432: Consider just returning the target string and then using `strings.Join` to combine them.

pkg/instrumentation/annotation.go
```diff
@@ -31,6 +31,21 @@ const (
...
 	annotationInjectNginxContainersName       = "instrumentation.opentelemetry.io/inject-nginx-container-names"
+	annotationJMXJVM                          = "cloudwatch.aws.amazon.com/inject-jmx-jvm"
+	annotationJMXTomcat                       = "cloudwatch.aws.amazon.com/inject-jmx-tomcat"
+	annotationJMXKafka                        = "cloudwatch.aws.amazon.com/inject-jmx-kafka"
+	annotationJMXKafkaConsumer                = "cloudwatch.aws.amazon.com/inject-jmx-kafka-consumer"
+	annotationJMXKafkaProducer                = "cloudwatch.aws.amazon.com/inject-jmx-kafka-producer"
+)
+
+var (
+	JMXAnnotationInjectionMap = map[string]string{
```
jefchien, line 42: nit: Do we need this map? Couldn't we trim the prefix to get the target system?

pkg/instrumentation/annotation.go
```diff
@@ -31,6 +31,21 @@ const (
 	annotationInjectApacheHttpdContainersName = "instrumentation.opentelemetry.io/apache-httpd-container-names"
 	annotationInjectNginx                     = "instrumentation.opentelemetry.io/inject-nginx"
 	annotationInjectNginxContainersName       = "instrumentation.opentelemetry.io/inject-nginx-container-names"
+	annotationJMXJVM                          = "cloudwatch.aws.amazon.com/inject-jmx-jvm"
+	annotationJMXTomcat                       = "cloudwatch.aws.amazon.com/inject-jmx-tomcat"
+	annotationJMXKafka                        = "cloudwatch.aws.amazon.com/inject-jmx-kafka"
+	annotationJMXKafkaConsumer                = "cloudwatch.aws.amazon.com/inject-jmx-kafka-consumer"
+	annotationJMXKafkaProducer                = "cloudwatch.aws.amazon.com/inject-jmx-kafka-producer"
```
jefchien, line 38: Consider keeping these in their own `const` block.

pkg/instrumentation/defaultinstrumentation.go
```diff
@@ -24,7 +24,7 @@ const (
 	httpsPrefix = "https://"
 )
 
-func getDefaultInstrumentation(agentConfig *adapters.CwaConfig) (*v1alpha1.Instrumentation, error) {
+func getDefaultInstrumentation(agentConfig *adapters.CwaConfig, extraJavaEnvVar map[string]string) (*v1alpha1.Instrumentation, error) {
```
jefchien, line 27: Does not need to be a language specific variable.

pkg/instrumentation/podmutator.go
```diff
@@ -232,7 +234,7 @@ func (pm *instPodMutator) Mutate(ctx context.Context, ns corev1.Namespace, pod c
 		pm.Recorder.Event(pod.DeepCopy(), "Warning", "InstrumentationRequestRejected", "support for Java auto instrumentation is not enabled")
 	}
 
-	if inst, err = pm.getInstrumentationInstance(ctx, ns, pod, annotationInjectNodeJS); err != nil {
+	if inst, err = pm.getInstrumentationInstance(ctx, ns, pod, annotationInjectNodeJS, make(map[string]string)); err != nil {
```
jefchien, line 237:
```suggestion
	if inst, err = pm.getInstrumentationInstance(ctx, ns, pod, annotationInjectNodeJS, nil); err != nil {
```

jefchien:
Tests are failing.

pkg/instrumentation/annotation.go
sethAmazon: I don't like doing string splits / postfix to get values. I think this is cleaner.

pkg/instrumentation/defaultinstrumentation.go
sethAmazon: Yes since. we have python and java. Why would these env var be in python.

pkg/instrumentation/defaultinstrumentation.go
lisguo: also this code path runs specifically when an Instrumentation CR is not found...if customers happen to have their own instrumentation CR for appsignals, these jmx env vars won't be injected. 

This means customers with a custom appsignals setup will have to manually add these vars to their Instrumentation spec. Is that expected?

pkg/instrumentation/annotation.go
```diff
@@ -31,6 +31,21 @@ const (
 	annotationInjectApacheHttpdContainersName = "instrumentation.opentelemetry.io/apache-httpd-container-names"
 	annotationInjectNginx                     = "instrumentation.opentelemetry.io/inject-nginx"
 	annotationInjectNginxContainersName       = "instrumentation.opentelemetry.io/inject-nginx-container-names"
+	annotationJMXJVM                          = "cloudwatch.aws.amazon.com/inject-jmx-jvm"
```
lisguo, line 34: have we tested a pod with just the jmx annotation and no opentelemetry instrumentation annotation?

sethAmazon:
The failing tests were due to gha being down not my changes.

pkg/instrumentation/annotation.go
sethAmazon: This is not in a e2e testable state only seeing the env vars for pods atm.

pkg/instrumentation/defaultinstrumentation.go
jefchien: We're passing it into each `getInstrumentationInstance`, which isn't language specific. If I wanted to set extra environment variables for python and go, would I need to add another map to this?

```go
getDefaultInstrumentation(agentConfig, nil, nil, map[string]string{"go-key": "go-value"})
```

Why not just use the same map? Each getInstrumentationInstance only gets the language specific instrumentation anyways so the other languages are ignored.

Or if you want to make the map language specific, make it a map of maps.

```
map[Type]map[string]string
```

The `annotation_type.go` file has a defined enum for java/python.

pkg/instrumentation/defaultinstrumentation.go
jefchien: How do we currently handle that case for appsignals?

pkg/instrumentation/annotation.go
jefchien: Or even in a separate file entirely.

pkg/instrumentation/podmutator.go
sethAmazon: I disagree. When we use nil we could have panics. I like the idea of empty objects over nulls.

pkg/instrumentation/annotation.go
sethAmazon: I'll make it another const block. I disagree with diff file, since all annotations should be here.

pkg/instrumentation/podmutator.go
jefchien: What are we doing with the map that would result in panics? It's pretty standard in go to pass around nil maps. Unless we're writing to it, it shouldn't panic.

https://go.dev/ref/spec#Map_types
> A nil map is equivalent to an empty map except that no elements may be added.

pkg/instrumentation/defaultinstrumentation.go
sethAmazon: I like map[Type]map[string]string

pkg/instrumentation/podmutator.go
sethAmazon: Again I do not like nil since it could lead to panic. Empty map is safer.

pkg/instrumentation/defaultinstrumentation.go
sethAmazon: I'm not understanding this. Mind if we huddle?

pkg/instrumentation/podmutator.go
sethAmazon: Guess I'm too java focused. Will change

pkg/instrumentation/podmutator.go
sethAmazon: done

pkg/instrumentation/podmutator.go
sethAmazon: done

sethAmazon:
"also this code path runs specifically when an Instrumentation CR is not found...if customers happen to have their own instrumentation CR for appsignals, these jmx env vars won't be injected.

This means customers with a custom appsignals setup will have to manually add these vars to their Instrumentation spec. Is that expected?"
when over this with @lisguo in a huddle. This is fine for this pr.

PR #187, opened 2024-06-17
Use kubectl wait to make the integration test more reliable

lisguo:
*Issue #, if available:* N/A

*Description of changes:* Instead of using sleep, we should use `kubectl wait` for resources to be available before continuing with the test


.github/workflows/operator-integration-test.yml
```diff
@@ -53,12 +53,14 @@ jobs:
       - name: Deploy operator to minikube
         run: |
           make deploy
+          kubectl wait --for=condition=Ready pod --all -n amazon-cloudwatch
```
lisguo, line 56: this is the main reason why it was flaky. operator wasn't ready to serve on the webhook.

Paramadon (approved):

PR #188, opened 2024-06-18
Fix command for java auto instrumentation on windows nodes

lisguo:
*Issue #, if available:* N/A

*Description of changes:*
- Fix command for java auto instrumentation on windows to use cmd.exe
- Modifies the injection logic to use the following command to copy the instrumentation jar:
```
Command:
      CMD
      /c
      copy
      javaagent.jar
      \otel-auto-instrumentation-java
```

The above will be injected for pods with the `kubernetes.io/os: windows` node selector. The EKS [documentation](https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html) requires this selector be added to pods.

---
**This will not enable support for windows. The operator still needs a windows build of the auto instrumentation image.**

Tested using a dev build on a windows node and saw java app running:
```
Defaulted container "sample-app" out of: sample-app, opentelemetry-auto-instrumentation-java (init)
Picked up JAVA_TOOL_OPTIONS:  -javaagent:/otel-auto-instrumentation-java/javaagent.jar
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
[otel.javaagent 2024-06-18 15:01:56:503 +0000] [main] INFO io.opentelemetry.javaagent.tooling.VersionLogger - opentelemetry-javaagent - version: 1.33.0-aws-SNAPSHOT
[otel.javaagent 2024-06-18 15:01:58:070 +0000] [main] INFO software.amazon.opentelemetry.javaagent.providers.AwsApplicationSignalsCustomizerProvider - AWS Application Signals enabled
```

### DIFF
--- a/pkg/instrumentation/javaagent.go
+++ b/pkg/instrumentation/javaagent.go
@@ -17,6 +17,11 @@ const (
 	javaInstrMountPath    = "/otel-auto-instrumentation-java"
 )
 
+var (
+	commandLinux   = []string{"cp", "/javaagent.jar", javaInstrMountPath + "/javaagent.jar"}
+	commandWindows = []string{"copy", "javaagent.jar", javaInstrMountPath}
+)
+
 func injectJavaagent(javaSpec v1alpha1.Java, pod corev1.Pod, index int) (corev1.Pod, error) {
 	// caller checks if there is at least one container.
 	container := &pod.Spec.Containers[index]
@@ -59,10 +64,15 @@ func injectJavaagent(javaSpec v1alpha1.Java, pod corev1.Pod, index int) (corev1.
 				},
 			}})
 
+		command := commandLinux
+		if pod.Spec.NodeSelector["kubernetes.io/os"] == "windows" {
+			command = commandWindows
+		}
+
 		pod.Spec.InitContainers = append(pod.Spec.InitContainers, corev1.Container{
 			Name:      javaInitContainerName,
 			Image:     javaSpec.Image,
-			Command:   []string{"cp", "/javaagent.jar", javaInstrMountPath + "/javaagent.jar"},
+			Command:   command,
 			Resources: javaSpec.Resources,
 			VolumeMounts: []corev1.VolumeMount{{
 				Name:      javaVolumeName,
--- a/pkg/instrumentation/javaagent.go
+++ b/pkg/instrumentation/javaagent.go
@@ -10,16 +10,17 @@ import (
 )
 
 const (
-	envJavaToolsOptions   = "JAVA_TOOL_OPTIONS"
-	javaJVMArgument       = " -javaagent:/otel-auto-instrumentation-java/javaagent.jar"
-	javaInitContainerName = initContainerName + "-java"
-	javaVolumeName        = volumeName + "-java"
-	javaInstrMountPath    = "/otel-auto-instrumentation-java"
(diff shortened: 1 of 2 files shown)


pkg/instrumentation/javaagent.go
```diff
@@ -10,11 +10,17 @@ import (
...
+	javaJVMArgument           = " -javaagent:/otel-auto-instrumentation-java/javaagent.jar"
+	javaInitContainerName     = initContainerName + "-java"
+	javaVolumeName            = volumeName + "-java"
+	javaInstrMountPath        = "/otel-auto-instrumentation-java"
+	javaInstrMountPathWindows = "\\otel-auto-instrumentation-java"
+)
+
+var (
+	commandLinux   = []string{"cp", "/javaagent.jar", javaInstrMountPath + "/javaagent.jar"}
+	commandWindows = []string{"CMD", "/c", "copy", "javaagent.jar", javaInstrMountPathWindows}
```
okankoAMZ, line 23: Just curious is there a reason why we use CMD /c instead of [Copy-Item](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.management/copy-item?view=powershell-7.4)?

okankoAMZ (approved):
just have a small question

pkg/instrumentation/javaagent.go
lisguo: windows nanoserver doesn't come with powershell so application pods wouldn't be able to run Copy-Item directly:

```
    Command:
      Copy-Item
      javaagent.jar
      \otel-auto-instrumentation-java
    State:          Waiting
      Reason:       RunContainerError
    Last State:     Terminated
      Reason:       StartError
      Message:      failed to start containerd task "a8f40a5dc034e4a566e577b42d0e86cd8d9e96ca70817c696b6a57bd5f32d39b": hcs::System::CreateProcess a8f40a5dc034e4a566e577b42d0e86cd8d9e96ca70817c696b6a57bd5f32d39b: The system cannot find the file specified.: unknown
      Exit Code:    128
...

Comment (lisguo) [reply pkg/instrumentation/javaagent.go:23]: Updated the description to not say powershell, but just cmd.exe

File: pkg/instrumentation/javaagent.go [hunk 2]
```diff
@@ -10,11 +10,17 @@ import (
...
-	javaJVMArgument       = " -javaagent:/otel-auto-instrumentation-java/javaagent.jar"
-	javaInitContainerName = initContainerName + "-java"
-	javaVolumeName        = volumeName + "-java"
-	javaInstrMountPath    = "/otel-auto-instrumentation-java"
+	envJavaToolsOptions       = "JAVA_TOOL_OPTIONS"
+	javaJVMArgument           = " -javaagent:/otel-auto-instrumentation-java/javaagent.jar"
+	javaInitContainerName     = initContainerName + "-java"
+	javaVolumeName            = volumeName + "-java"
+	javaInstrMountPath        = "/otel-auto-instrumentation-java"
+	javaInstrMountPathWindows = "\\otel-auto-instrumentation-java"
```
movence, line 18: do we need a test case for this?

lisguo: Test case was added in `func TestInjectJavaagentWindows`

movence (approved):

PR #203, opened 2024-08-05
Change test failure to echo a warning when checking k8s version dependencies

lisguo:
*Issue #, if available:*

*Description of changes:*
Upstream opentelemetry-operator does not always ensure k8s dependencies are up to date when supporting a new version of k8s. Our current compatibility model depends on upstream, so this check is a blocker for future versions.

Example run: https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/10356859832

### DIFF
--- a/.github/workflows/eks-add-on-integ-test.yml
+++ b/.github/workflows/eks-add-on-integ-test.yml
@@ -94,7 +94,6 @@ jobs:
           echo or k8s.io/component-base $(go list -m k8s.io/component-base) is less than ${{ matrix.arrays.k8sVersion }}
           echo or k8s.io/kubectl $(go list -m k8s.io/kubectl) is less than ${{ matrix.arrays.k8sVersion }}
           echo "[WARNING] Kubernetes dependencies are lower than the latest Kubernetes cluster version. Please check changelog to ensure no breaking changes: https://kubernetes.io/releases/"
-          exit 1;
           fi
 
       - name: Verify Terraform version
--- a/.github/workflows/eks-add-on-integ-test.yml
+++ b/.github/workflows/eks-add-on-integ-test.yml
@@ -81,21 +81,22 @@ jobs:
           role-to-assume: ${{ env.TERRAFORM_AWS_ASSUME_ROLE }}
           aws-region: us-west-2
 
-      - name: Confirm EKS Version Support
-        run: |
-          if [[ 
-            $(go list -m k8s.io/client-go | cut -d ' ' -f 2 | cut -d '.' -f 2) -lt $( echo ${{ matrix.arrays.k8sVersion }} | cut -d '.' -f 2)
-            || $(go list -m k8s.io/apimachinery | cut -d ' ' -f 2 | cut -d '.' -f 2) -lt $( echo ${{ matrix.arrays.k8sVersion }} | cut -d '.' -f 2)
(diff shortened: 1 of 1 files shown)


.github/workflows/eks-add-on-integ-test.yml
```diff
@@ -81,21 +81,22 @@ jobs:
...
-            || $(go list -m k8s.io/kubectl | cut -d ' ' -f 2 | cut -d '.' -f 2) -lt $( echo ${{ matrix.arrays.k8sVersion }} | cut -d '.' -f 2)
-          ]]; then 
-          echo k8s.io/client-go $(go list -m k8s.io/client-go) is less than ${{ matrix.arrays.k8sVersion }}
-          echo or k8s.io/apimachinery $(go list -m k8s.io/apimachinery) is less than ${{ matrix.arrays.k8sVersion }}
-          echo or k8s.io/component-base $(go list -m k8s.io/component-base) is less than ${{ matrix.arrays.k8sVersion }}
-          echo or k8s.io/kubectl $(go list -m k8s.io/kubectl) is less than ${{ matrix.arrays.k8sVersion }}, fail test
-          echo "please run go get -u && go mod tidy"
-          exit 1; 
-          fi
+#      Upstream opentelemetry-operator does not comply with k8s version dependencies for new k8s version support
```
mitali-salvi, line 84: Why is this commented out ? Are we disabling this test ?

.github/workflows/eks-add-on-integ-test.yml
```diff
@@ -83,18 +83,17 @@ jobs:
...
             || $(go list -m k8s.io/component-base | cut -d ' ' -f 2 | cut -d '.' -f 2) -lt $( echo ${{ matrix.arrays.k8sVersion }} | cut -d '.' -f 2)
             || $(go list -m k8s.io/kubectl | cut -d ' ' -f 2 | cut -d '.' -f 2) -lt $( echo ${{ matrix.arrays.k8sVersion }} | cut -d '.' -f 2)
-          ]]; then 
+          ]]; then
           echo k8s.io/client-go $(go list -m k8s.io/client-go) is less than ${{ matrix.arrays.k8sVersion }}
           echo or k8s.io/apimachinery $(go list -m k8s.io/apimachinery) is less than ${{ matrix.arrays.k8sVersion }}
           echo or k8s.io/component-base $(go list -m k8s.io/component-base) is less than ${{ matrix.arrays.k8sVersion }}
-          echo or k8s.io/kubectl $(go list -m k8s.io/kubectl) is less than ${{ matrix.arrays.k8sVersion }}, fail test
-          echo "please run go get -u && go mod tidy"
-          exit 1;
```
mitali-salvi, line 97: Cam we add an always() tag to the terraform verification step below so the workflow keeps running and the ```Confirm EKS Version Support``` step result is visible (red/green). I feel just logging the error will just to ignoring the test since it will always show as successful.

```
jobs:
  job1:
  job2:
    needs: job1
  job3:
    if: always()
    needs: [job1, job2]
```
https://stackoverflow.com/a/58859404

lisguo: Hm so I tried adding `always()` to the subsequent steps to allow this step to fail but it then fails the entire workflow: https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/10357729837

mitali-salvi (approved):

sky333999 (approved):

mitali-salvi (approved):

PR #204, opened 2024-08-15
Update manager rbac cluster role

lisguo:
*Issue #, if available:*

*Description of changes:*
Similar to the helm chart PR: https://github.com/aws-observability/helm-charts/pull/83

This is to update the manager role in the operator so that we can test against the new privilege changes

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -4,214 +4,54 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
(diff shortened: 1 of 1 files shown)


mitali-salvi (commented):
Can we update the kubebuilder annotations instead ? https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/controllers/amazoncloudwatchagent_controller.go#L66

PR #245, opened 2024-10-09
Add integration tests for JMX.

musa-asad:
### Description of changes
Adds integration tests to check if enabling annotations adds expected JMX environmental variables for `app_signals` and without.

### Testing
`make test`
Screenshot 2024-10-09 at 9 40 03 AM

See passing tests in `Checks`: https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/11283409192?pr=245.

----------------


.github/workflows/operator-integration-test.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           kubectl wait --for=condition=Ready pod --all -n default
           kubectl get pods -A 
           kubectl describe pods -n default
-          go run integration-tests/manifests/cmd/validate_instrumentation_vars.go default integration-tests/java/default_instrumentation_java_env_variables.json
+          go run integration-tests/manifests/cmd/validate_instrumentation_vars.go default integration-tests/java/default_instrumentation_java_env_variables.json app_signals
```
Paramadon, line 81: If your passing app_signals as an argument evertime, what is the point of the argument? Is there a case for not_app_signals? And if so, do you change the agent configuration to reflect that on the env vars?

integration-tests/manifests/cmd/validate_instrumentation_vars.go
```diff
@@ -89,6 +99,15 @@ func verifyInstrumentationEnvVariables(clientset *kubernetes.Clientset, namespac
 	}
 	fmt.Println("JSON data:", jsonData)
 
+	if appSignals == "no_app_signals" {
```
Paramadon, line 102: Does your test ever go within this if statement?

.github/workflows/operator-integration-test.yml
musa-asad: Yeah, I had to add it so it wouldn't cause a missing argument failure. The one time it is used is when we're checking for `no_app_signals`. Then it runs the check to make sure application signals environmental variables aren't present. However, I can adjust the logic to check the agent configuration itself.

integration-tests/manifests/cmd/validate_instrumentation_vars.go
```diff
@@ -51,7 +61,7 @@ func main() {
 	}
 }
 
-func verifyInstrumentationEnvVariables(clientset *kubernetes.Clientset, namespace, jsonPath string) bool {
+func verifyInstrumentationEnvVariables(clientset *kubernetes.Clientset, namespace, jsonPath string, appSignals string) bool {
```
Paramadon, line 64: Instead of passing an argument can we look at the agent config instead for checking if app signals is enabled?

integration-tests/manifests/cmd/validate_instrumentation_vars.go
musa-asad: I can add a log for that.

integration-tests/manifests/cloudwatch-agent-daemonset-no-app-signals.yaml
```diff
@@ -0,0 +1,70 @@
+apiVersion: cloudwatch.aws.amazon.com/v1alpha1
```
Paramadon, line 1: Oh I see, you change the daemonset here

.github/workflows/operator-integration-test.yml
```diff
@@ -165,9 +164,69 @@ jobs:
...
+        run: |
+          kubectl delete pods --all -n default
+          sleep 5
+          cat integration-tests/jmx/sample-deployment-jmx.yaml
+          kubectl apply -f integration-tests/jmx/sample-deployment-jmx.yaml
+          sleep 5
+          kubectl wait --for=condition=Available deployment/nginx -n default
+          kubectl get pods -A 
+          kubectl describe pods -n default
+          go run integration-tests/manifests/cmd/validate_instrumentation_vars.go default integration-tests/jmx/default_instrumentation_jmx_env_variables_no_app_signals.json no_app_signals
```
Paramadon, line 219: Ahh okay I see this is the no_app_signals case and you use a json without app_signals.

PR #254, opened 2024-10-21
Fix service name source being added multiple times for multiple injections

nathalapooja:
*Issue #, if available:*
With the current code, if the customer pod has multiple injections , the service name source is being added multiple times and with invalid value from the second injection.
```
OTEL_RESOURCE_ATTRIBUTES: com.amazonaws.cloudwatch.entity.internal.service.name.source=K8sWorkload,k8s.container.name=petclinic,k8s.deployment.name=petclinic-1,k8s.namespace.name=default,k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME),k8s.replicaset.name=petclinic-1-6bb849b4d5,service.version=latest,**com.amazonaws.cloudwatch.entity.internal.service.name.source=Instrumentation,k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME)**
```

*Description of changes:*
added code to check the existing OTEL resource attributes and add the attribute only when the existing value is NULL

Tested using `Unit Tests`

Manual Testing on petclinic pod without custom service name

Before
```
OTEL_RESOURCE_ATTRIBUTES:   k8s.container.name=petclinic-container,k8s.daemonset.name=petclinic-pod,k8s.namespace.name=default,k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME),service.vers │
│ ion=latest,com.amazonaws.cloudwatch.entity.internal.service.name.source=K8sWorkload,k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME),com.amazonaws.cloudwatch.entity.internal.service.name.source=Instrumentation

```

After
```
OTEL_RESOURCE_ATTRIBUTES:                        com.amazonaws.cloudwatch.entity.internal.service.name.source=K8sWorkload,k8s.container.name=petclinic-container,k8s.daemonset.name=petclinic-pod,k8s.namespace.name=default,k8s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_ │
│ NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME),service.version=latest   
```

With Custom service name:

After
```
OTEL_RESOURCE_ATTRIBUTES:                        deployment.environment=compass-env,com.amazonaws.cloudwatch.entity.internal.service.name.source=Instrumentation,k8s.container.name=petclinic-container,k8s.daemonset.name=petclinic-pod,k8s.namespace.name=default,k8 │
│ s.node.name=$(OTEL_RESOURCE_ATTRIBUTES_NODE_NAME),k8s.pod.name=$(OTEL_RESOURCE_ATTRIBUTES_POD_NAME),service.version=latest  
```

### DIFF
--- a/pkg/instrumentation/sdk.go
+++ b/pkg/instrumentation/sdk.go
@@ -160,7 +160,8 @@ func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations
 			index := getContainerIndex(container, pod)
 			// Apache agent is configured via config files rather than env vars.
 			// Therefore, service name, otlp endpoint and other attributes are passed to the agent injection method
-			pod = injectApacheHttpdagent(i.logger, otelinst.Spec.ApacheHttpd, pod, index, otelinst.Spec.Endpoint, i.createResourceMap(ctx, otelinst, ns, pod, index))
+			resMap, _ := i.createResourceMap(ctx, otelinst, ns, pod, index)
+			pod = injectApacheHttpdagent(i.logger, otelinst.Spec.ApacheHttpd, pod, index, otelinst.Spec.Endpoint, resMap)
 			pod = i.injectCommonEnvVar(otelinst, pod, index)
 			pod = i.injectCommonSDKConfig(ctx, otelinst, ns, pod, index, index)
 			pod = i.setInitContainerSecurityContext(pod, pod.Spec.Containers[index].SecurityContext, apacheAgentInitContainerName)
@@ -178,7 +179,8 @@ func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations
 			index := getContainerIndex(container, pod)
 			// Nginx agent is configured via config files rather than env vars.
 			// Therefore, service name, otlp endpoint and other attributes are passed to the agent injection method
-			pod = injectNginxSDK(i.logger, otelinst.Spec.Nginx, pod, index, otelinst.Spec.Endpoint, i.createResourceMap(ctx, otelinst, ns, pod, index))
+			resMap, _ := i.createResourceMap(ctx, otelinst, ns, pod, index)
+			pod = injectNginxSDK(i.logger, otelinst.Spec.Nginx, pod, index, otelinst.Spec.Endpoint, resMap)
 			pod = i.injectCommonEnvVar(otelinst, pod, index)
 			pod = i.injectCommonSDKConfig(ctx, otelinst, ns, pod, index, index)
 		}
@@ -243,7 +245,7 @@ func (i *sdkInjector) injectCommonEnvVar(otelinst v1alpha1.Instrumentation, pod 
 // and appIndex should represent the application being instrumented.
 func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alpha1.Instrumentation, ns corev1.Namespace, pod corev1.Pod, agentIndex int, appIndex int) corev1.Pod {
 	container := &pod.Spec.Containers[agentIndex]
-	resourceMap := i.createResourceMap(ctx, otelinst, ns, pod, appIndex)
+	resourceMap, existingRes := i.createResourceMap(ctx, otelinst, ns, pod, appIndex)
 	idx := getIndexOfEnv(container.Env, constants.EnvOTELServiceName)
 	serviceNameSource := constants.SourceInstrumentation
 	if idx == -1 {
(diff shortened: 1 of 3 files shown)


pkg/instrumentation/sdk.go
```diff
@@ -263,8 +263,22 @@ func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alph
 		}
 	}
 
+	// get existing resources env var and parse it into a map
+	existingRes := map[string]string{}
+	existingResourceEnvIdx := getIndexOfEnv(pod.Spec.Containers[appIndex].Env, constants.EnvOTELResourceAttrs)
```
sky333999, line 268: Looks like `createResourceMap` was already doing this and reading the env var to put it in `resourceMap`. So why wasnt that check i.e. `resourceMap[key] == ""` not sufficient for things like K8SPodNameKey?

nathalapooja: for k, v := range otelinst.Spec.Resource.Attributes {
		if !existingRes[k] {
			res[k] = v
		}
	}
If the attribute doesn't exist in existingRes, then the res will be updated. So when the second injection happens, res doesn't have the semconv.K8SPodNameKey key in it as it was existing value. So just checking the resourceMap[string(semconv.K8SPodNameKey)] == "" will be true and the key is added again

lisguo, line 268: seems like this code borrows the first few lines of `createResourceMap`...can we prevent duplicate code by creating a helper function for the following:

```
// get existing resources env var and parse it into a map
	existingRes := map[string]bool{}
	existingResourceEnvIdx := getIndexOfEnv(pod.Spec.Containers[index].Env, constants.EnvOTELResourceAttrs)
	if existingResourceEnvIdx > -1 {
		existingResArr := strings.Split(pod.Spec.Containers[index].Env[existingResourceEnvIdx].Value, ",")
		for _, kv := range existingResArr {
			keyValueArr := strings.Split(strings.TrimSpace(kv), "=")
			if len(keyValueArr) != 2 {
				continue
			}
			existingRes[keyValueArr[0]] = true
		}
	}

	res := map[string]string{}
	for k, v := range otelinst.Spec.Resource.Attributes {
		if !existingRes[k] {
			res[k] = v
		}
	}
```

and then using this in `createResourceMap`?

pkg/instrumentation/sdk.go
```diff
@@ -263,8 +263,22 @@ func (i *sdkInjector) injectCommonSDKConfig(ctx context.Context, otelinst v1alph
...
+			if len(keyValueArr) != 2 {
+				continue
+			}
+			existingRes[keyValueArr[0]] = keyValueArr[1]
+		}
+	}
+
 	// Some attributes might be empty, we should get them via k8s downward API
-	if resourceMap[string(semconv.K8SPodNameKey)] == "" {
+	if existingRes[string(semconv.K8SPodNameKey)] == "" && resourceMap[string(semconv.K8SPodNameKey)] == "" {
```
lisguo, line 281: Not sure I understand...why isn't `resourceMap` sufficient to determine if the k8s pod name key is missing?

pkg/instrumentation/sdk.go
nathalapooja: We are using map[string]bool in createResourceMap and in the new code we use map[string]string{}.

pkg/instrumentation/sdk.go
nathalapooja: for k, v := range otelinst.Spec.Resource.Attributes {
	if !existingRes[k] {
		res[k] = v
	}
}
If the attribute doesn't exist in existingRes, then the res will be updated. So when the second injection happens, res doesn't have the semconv.K8SPodNameKey key in it as it was existing value. So just checking the resourceMap[string(semconv.K8SPodNameKey)] == "" will be true and the key is added again

lisguo (approved):

PR #255, opened 2024-10-21
Clean up  managed resources when disabled

okankoAMZ:
## *Issue #, if available:*
Currently when the collector changes modes or target allocator is toggled, the reconciler will create the new resource. However, reconciler will not remove previous resources.  This PR's goal is to let reconciler remove un-used managed resources. 
## *Description of changes:*
* Added a new reconcile function that will delete resources that are not built (required new function to not effect dcgm and other controllers)
* Added a managed resource finder that tracks: 
- ConfigMaps
- DaemonSets
- Deployments
- Services
- ServiceAccounts
- StatefulSets
## Testing 
I first deploy with TA enabled then disable it via helm charts.
### When target allocator is enabled: 
<details>
<summary>Pods</summary>

Screenshot 2024-10-22 at 11 47 09 AM
</details>
<details>
<summary>Deployments</summary>

Screenshot 2024-10-22 at 11 47 38 AM

</details>
<details>
<summary>ConfigMaps</summary>

Screenshot 2024-10-22 at 11 47 52 AM

</details>
<details>
<summary>Services</summary>

Screenshot 2024-10-22 at 11 48 08 AM

</details>
<details>
<summary>ServiceAccounts</summary>

Screenshot 2024-10-22 at 11 48 21 AM

</details>

### When target allocator is disabled: 
<details>
<summary>Pods</summary>

Screenshot 2024-10-22 at 11 49 51 AM

</details>
<details>
<summary>Deployments</summary>

Screenshot 2024-10-22 at 11 50 00 AM

</details>
<details>
<summary>ConfigMaps</summary>

Screenshot 2024-10-22 at 11 50 09 AM

</details>
<details>
<summary>Services</summary>

Screenshot 2024-10-22 at 11 50 26 AM

</details>
<details>
<summary>ServiceAccounts</summary>

Screenshot 2024-10-22 at 11 50 39 AM

</details>

### DIFF
--- a/controllers/amazoncloudwatchagent_controller.go
+++ b/controllers/amazoncloudwatchagent_controller.go
@@ -6,15 +6,14 @@ package controllers
 
 import (
 	"context"
-	"github.com/aws/amazon-cloudwatch-agent-operator/internal/manifests/manifestutils"
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -22,6 +21,7 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent-operator/apis/v1alpha1"
 	"github.com/aws/amazon-cloudwatch-agent-operator/internal/config"
 	"github.com/aws/amazon-cloudwatch-agent-operator/internal/manifests"
+	"github.com/aws/amazon-cloudwatch-agent-operator/internal/manifests/manifestutils"
 	collectorStatus "github.com/aws/amazon-cloudwatch-agent-operator/internal/status/collector"
 )
 
--- a/controllers/amazoncloudwatchagent_controller.go
+++ b/controllers/amazoncloudwatchagent_controller.go
@@ -46,8 +46,7 @@ type Params struct {
 func (r *AmazonCloudWatchAgentReconciler) findCloudWatchAgentOwnedObjects(ctx context.Context, owner v1alpha1.AmazonCloudWatchAgent) (map[types.UID]client.Object, error) {
 	// Define a map to store the owned objects
 	ownedObjects := make(map[types.UID]client.Object)
-	selector := manifestutils.SelectorLabels(owner.ObjectMeta, "*")
-	delete(selector, "app.kubernetes.io/component") //ignore components
+	selector := manifestutils.SelectorLabelsForAllOperatorManaged(owner.ObjectMeta)
 	listOps := &client.ListOptions{
 		Namespace:     owner.Namespace,
 		LabelSelector: labels.SelectorFromSet(selector),
--- a/controllers/amazoncloudwatchagent_controller.go
+++ b/controllers/amazoncloudwatchagent_controller.go
@@ -6,6 +6,9 @@ package controllers
 
 import (
 	"context"
+	"github.com/aws/amazon-cloudwatch-agent-operator/internal/manifests/manifestutils"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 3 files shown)


controllers/amazoncloudwatchagent_controller.go
```diff
@@ -40,6 +43,80 @@ type Params struct {
...
+	serviceAccountList := &corev1.ServiceAccountList{}
+	deploymentList := &appsv1.DeploymentList{}
+	statefulSetList := &appsv1.StatefulSetList{}
+	daemonSetList := &appsv1.DaemonSetList{}
+	var err error
+
+	// List ConfigMaps
+	err = r.List(ctx, configMapList, listOps)
+	if err != nil {
+		return nil, err
```
musa-asad, line 67: Can you specify the resource it failed to list?
```suggestion
		return nil, fmt.Errorf("failed to list ConfigMaps: %w", err)
```

controllers/amazoncloudwatchagent_controller.go
```diff
@@ -40,6 +43,80 @@ type Params struct {
 	Config   config.Config
 }
 
+func (r *AmazonCloudWatchAgentReconciler) findCloudWatchAgentOwnedObjects(ctx context.Context, owner v1alpha1.AmazonCloudWatchAgent) (map[types.UID]client.Object, error) {
```
musa-asad, line 46: Can you set up a unit test for this function with a sample `v1alpha1.AmazonCloudWatchAgent` instance and see if this properly detects the created resources?

controllers/amazoncloudwatchagent_controller.go
```diff
@@ -40,6 +43,80 @@ type Params struct {
...
+	}
+
+	// List DaemonSets
+	err = r.List(ctx, daemonSetList, listOps)
+	if err != nil {
+		return nil, err
+	}
+	for i := range daemonSetList.Items {
+		ownedObjects[daemonSetList.Items[i].GetUID()] = &daemonSetList.Items[i]
+	}
```
musa-asad, line 115: Would it make sense to create a helper function for each of these resources since it uses similar logic?

controllers/common.go
```diff
@@ -57,6 +58,74 @@ func BuildCollector(params manifests.Params) ([]client.Object, error) {
...
+func reconcileDesiredObjectsWPrune(ctx context.Context, kubeClient client.Client, logger logr.Logger, owner v1alpha1.AmazonCloudWatchAgent, scheme *runtime.Scheme,
+	desiredObjects []client.Object,
+	searchOwnedObjectsFunc func(ctx context.Context, owner v1alpha1.AmazonCloudWatchAgent) (map[types.UID]client.Object, error),
+) error {
+	var errs []error
+	existingObjectMap := make(map[types.UID]client.Object)
+	existingObjectList := []client.Object{}
+	previouslyOwnedObjects, err := searchOwnedObjectsFunc(ctx, owner)
+	if err != nil {
+		return err
```
musa-asad, line 71: Can we improve on the error message here? I'm not sure what it would look like from a customer's perspective.

controllers/common.go
```diff
@@ -57,6 +58,74 @@ func BuildCollector(params manifests.Params) ([]client.Object, error) {
 	return resources, nil
 }
 
+// @TODO: Later merge this in to reconcileDesiredObjects after implementing search functions for DCGM
+func reconcileDesiredObjectsWPrune(ctx context.Context, kubeClient client.Client, logger logr.Logger, owner v1alpha1.AmazonCloudWatchAgent, scheme *runtime.Scheme,
```
musa-asad, line 62: Likewise, would it be plausible to include this function in a unit test to see if it appropriately deletes?

controllers/amazoncloudwatchagent_controller.go
okankoAMZ: It already does do that : 
Screenshot 2024-10-22 at 2 47 37 PM

controllers/amazoncloudwatchagent_controller.go
okankoAMZ: There are no unit-test available for controllers, they are tested via integration tests

controllers/common.go
okankoAMZ: There are no unit-test available for controllers, they are tested via integration tests

controllers/amazoncloudwatchagent_controller.go
okankoAMZ: Because they are interfaces you cannot access Items without this method. We do something similar to this in auto instrumentation here: https://github.com/aws/amazon-cloudwatch-agent-operator/blob/74fb8a174973a17dda77fdfafbfaeff46b7ea9c1/pkg/instrumentation/auto/annotation.go#L69-L83 

I choose to use blocks instead of a switch because I find the error handling easier to follow.

musa-asad (approved):

controllers/amazoncloudwatchagent_controller.go
```diff
@@ -40,6 +43,80 @@ type Params struct {
 	Config   config.Config
 }
 
+func (r *AmazonCloudWatchAgentReconciler) findCloudWatchAgentOwnedObjects(ctx context.Context, owner v1alpha1.AmazonCloudWatchAgent) (map[types.UID]client.Object, error) {
+	// Define a map to store the owned objects
+	ownedObjects := make(map[types.UID]client.Object)
+	selector := manifestutils.SelectorLabels(owner.ObjectMeta, "*")
```
jefchien, line 49: nit: Change the "*" to an empty string to make it clear that we're not using it. Without knowing what this function does, I would assume "*" selects everything.

controllers/common.go
```diff
@@ -57,6 +58,74 @@ func BuildCollector(params manifests.Params) ([]client.Object, error) {
...
+		}
+
+		l.V(1).Info(fmt.Sprintf("desired has been %s", op))
+	}
+	if len(errs) > 0 {
+		return fmt.Errorf("failed to create objects for %s: %w", owner.GetName(), errors.Join(errs...))
+	}
+
+	for _, obj := range existingObjectList {
+		existingObjectMap[obj.GetUID()] = obj.DeepCopyObject().(client.Object)
```
jefchien, line 119: Do we need to do another DeepCopy? We're already making one on line 89.

controllers/amazoncloudwatchagent_controller.go
```diff
@@ -40,6 +43,80 @@ type Params struct {
...
+	delete(selector, "app.kubernetes.io/component") //ignore components
```
jefchien, line 50: nit: It's unlikely to change, but could we add a const for this key in `manifestutils` and use that instead?

controllers/common.go
jefchien, line 62: nit: This is almost entirely the same function as `reconcileDesiredObjects`. The main difference is that it collects the "existing" `client.Object`s. Could we just make a helper function that does this for us and have the `reconcileDesiredObjects` function call that and ignore the slice?

```go
func helper(ctx context.Context, kubeClient client.Client, logger logr.Logger, owner metav1.Object, scheme *runtime.Scheme, desiredObjects ...client.Object) ([]client.Object, error)
```

Main concern is that if we pull in changes, we're likely only going to remember to update the `reconcileDesiredObjects` function, which is in upstream, and forget to update this one.

controllers/common.go
```diff
@@ -104,6 +173,29 @@ func reconcileDesiredObjects(ctx context.Context, kubeClient client.Client, logg
 	return nil
 }
 
+func pruneStaleObjects(ctx context.Context, kubeClient client.Client, logger logr.Logger, previouslyOwnedMap, desiredMap map[types.UID]client.Object) error {
+	// Pruning owned objects in the cluster which are not should not be present after the reconciliation.
```
jefchien, line 177: nit:
```suggestion
	// Pruning owned objects in the cluster which should not be present after the reconciliation.
```

musa-asad (approved):

jefchien (approved):

musa-asad (approved):

PR #250, opened 2024-10-16
Add integration tests for JMX.

musa-asad:
### Description of changes
Adds integration tests to check if enabling annotations adds expected JMX environmental variables for `app_signals` and without.

### Testing
`make test`
Screenshot 2024-10-09 at 9 40 03 AM

See passing tests in `Checks`.

Addressed @Paramadon's comments from https://github.com/aws/amazon-cloudwatch-agent-operator/pull/245/.

----------------
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #257, opened 2024-10-23
Update sample deployment to include NodeJS

musa-asad:
*Description of changes:*
We want to inject all language annotations to our `sample-deployment-all-languages.yaml`. This change makes sure to include `nodejs`.


musa-asad:
Added to #250.

musa-asad:
Added to #250.

PR #258, opened 2024-10-23
Injecting Prometheus path if not specified in agent config

Paramadon:
### Description of issue
Currently there is a bug for prometheus eks on CWA where if a customer does not specify the config file path in the prometheus section it is not auto injected in the agent.json. This pr fixes that and sets the prometheus_config_path to `/etc/prometheusconfig/prometheus.yaml`

*Description of changes:*

### Image of user not specifying `prometheus_config_path` in their config and the resulting agent.json (config file path is at the end)
Screenshot 2024-10-22 at 8 57 12 PM
Screenshot 2024-10-22 at 8 58 04 PM

### Image of user specifying `prometheus_config_path` in their config and the resulting agent.json
Screenshot 2024-10-22 at 8 57 04 PM
Screenshot 2024-10-22 at 8 56 20 PM

### Testing
Added unit test for both cases where there is not `prometheus_config_path` and when there is.

### Steps to replicate and the solution
1. Clone operator repo and checkout injectingPromPath
2. Run `make container` and then push that image to an ecr repo 
3. Clone helm chart repo and checkout ta-helm branch
4. Replace the manager image with the one you pushed to the ecr repo. 
5. You can change the prometheus_config_path under the agent config to test out if the fix is working. 
6. Run `helm upgrade --install amazon-cloudwatch-observability ../helm-charts/charts/amazon-cloudwatch-observability --values ../helm-charts/charts/amazon-cloudwatch-observability/values.yaml --set clusterName=<cluster-nam> --set region=<region> --namespace amazon-cloudwatch --create-namespace` 
7. If that fails just delete the amazon-cloudwatch namespace and try again.
8. Now you can check the `cloudwatch-agent-w-prom` config map and that it has the agent json with the correct prometheus_config_path.

### Ran govulcheck ./..
Screenshot 2024-10-28 at 5 11 29 PM


musa-asad:
You shouldn't need to edit the `Dockerfile`, `go.mod`, and `go.sum`.

internal/manifests/collector/config_replace_test.go
```diff
@@ -358,7 +359,7 @@ func TestReplacePrometheusConfig(t *testing.T) {
     scrape_interval: 1m
     scrape_timeout: 10s
 target_allocator:
-  endpoint: http://target-allocator-service:80
+  endpoint: https://target-allocator-service:80
```
musa-asad, line 362: Thanks for fixing these.

internal/manifests/collector/config_replace.go
```diff
@@ -33,12 +34,38 @@ type Config struct {
 }
 
 func ReplaceConfig(instance v1alpha1.AmazonCloudWatchAgent) (string, error) {
+	// Parse the original configuration from instance.Spec.Config
 	config, err := adapters.ConfigFromJSONString(instance.Spec.Config)
 	if err != nil {
 		return "", err
 	}
+	fmt.Println("Original config:", config)
```
musa-asad, line 42: nit: spacing
```suggestion
	fmt.Println("Original config: ", config)
```

internal/manifests/collector/config_replace.go
```diff
@@ -33,12 +34,38 @@ type Config struct {
...
 
-	out, err := json.Marshal(config)
+	conf := confmap.NewFromStringMap(config)
+
+	prometheusFilePath := conf.Get("logs::metrics_collected::prometheus::prometheus_config_path")
```
musa-asad, line 46: 👌

internal/manifests/collector/config_replace.go
```diff
@@ -33,12 +34,38 @@ type Config struct {
...
+			"metrics_collected": map[string]interface{}{
+				"prometheus": map[string]interface{}{
+					"prometheus_config_path": prometheusFilePath,
+				},
+			},
+		},
+	})
+
+	// Merge the Prometheus configuration
+	err = conf.Merge(prometheusConfig)
```
musa-asad, line 62: 👌

internal/manifests/collector/config_replace.go
```diff
@@ -33,12 +34,38 @@ type Config struct {
...
 	}
+	fmt.Println("Original config:", config)
 
-	out, err := json.Marshal(config)
+	conf := confmap.NewFromStringMap(config)
+
+	prometheusFilePath := conf.Get("logs::metrics_collected::prometheus::prometheus_config_path")
+	if prometheusFilePath == nil {
+		prometheusFilePath = "/etc/prometheusconfig/prometheus.yaml"
+	}
```
musa-asad, line 49: Nice way to fetch the `prometheus_config_path`, but do we need to do this? The customer should always be referencing the `prometheus.yaml` stored in the volume mount.

internal/manifests/collector/config_replace.go
```diff
@@ -33,12 +34,38 @@ type Config struct {
...
+
+	prometheusConfig := confmap.NewFromStringMap(map[string]interface{}{
+		"logs": map[string]interface{}{
+			"metrics_collected": map[string]interface{}{
+				"prometheus": map[string]interface{}{
+					"prometheus_config_path": prometheusFilePath,
+				},
+			},
+		},
+	})
```
musa-asad, line 59: Nice way to fetch the `prometheus_config_path`, but do we need to do this? The customer should always be referencing the `prometheus.yaml` stored in the volume mount, right?
```suggestion
	prometheusConfig := confmap.NewFromStringMap(map[string]interface{}{
		"logs": map[string]interface{}{
			"metrics_collected": map[string]interface{}{
				"prometheus": map[string]interface{}{
					"prometheus_config_path": "/etc/prometheusconfig/prometheus.yaml",
				},
			},
		},
	})
```

Paramadon: A customer can have a different prometheus config path in their agent json so I don't think we should just override that. We will default to "/etc/prometheusconfig/prometheus.yaml" if there is not config path specified.

musa-asad: Also, it looks like we do use this path under "metrics": https://github.com/aws/amazon-cloudwatch-agent/pull/1386. So we may need to implement that.

internal/manifests/collector/config_replace.go
```diff
@@ -33,12 +34,57 @@ type Config struct {
...
 func ReplaceConfig(instance v1alpha1.AmazonCloudWatchAgent) (string, error) {
+	// Parse the original configuration from instance.Spec.Config
 	config, err := adapters.ConfigFromJSONString(instance.Spec.Config)
 	if err != nil {
 		return "", err
 	}
 
-	out, err := json.Marshal(config)
+	conf := confmap.NewFromStringMap(config)
+
```
musa-asad, line 44: You would also need this check because if it is empty, then the `/etc/prometheusconfig/prometheus.yaml` won't be created.
```go
!instance.Spec.Prometheus.IsEmpty()
```

internal/manifests/collector/config_replace.go
Paramadon: Changed it so that we only add prometheus_config_path if prometheus section exist we make sure that configmap for prometheus was created.

internal/manifests/collector/config_replace.go
Paramadon: done

musa-asad (dismissed):
Good testing.

okankoAMZ (changes requested):
Cannot build due to: 
```
imported by a module that requires go 1.21
```

internal/manifests/collector/config_replace.go
```diff
@@ -33,12 +34,57 @@ type Config struct {
...
+	prometheusFilePath := conf.Get("logs::metrics_collected::prometheus::prometheus_config_path")
+	if prometheusFilePath == nil {
+		prometheusFilePath = "/etc/prometheusconfig/prometheus.yaml"
+	}
+	if conf.IsSet("logs::metrics_collected::prometheus") && !instance.Spec.Prometheus.IsEmpty() {
+		prometheusConfig := confmap.NewFromStringMap(map[string]interface{}{
```
okankoAMZ, line 50: nit: this could be a global variable called `DefaultPrometheusPath` and can be reused. We don't need to create a new map everytime.

internal/manifests/collector/config_replace.go
```diff
@@ -33,12 +34,57 @@ type Config struct {
...
+		err = conf.Merge(prometheusConfig)
+		if err != nil {
+			return "", err
+		}
+	}
+	prometheusFilePath = conf.Get("metrics::metrics_collected::prometheus::prometheus_config_path")
+	if prometheusFilePath == nil {
+		prometheusFilePath = "/etc/prometheusconfig/prometheus.yaml"
+	}
+	if conf.IsSet("metrics::metrics_collected::prometheus") && !instance.Spec.Prometheus.IsEmpty() {
```
okankoAMZ, line 69: nit: this could be a global variable called DefaultMetricsPrometheusPath and can be reused. We don't need to create a new map everytime.

okankoAMZ (dismissed):
Looks good, just want to point out that this would mean, the customer still has to define prometheus section to enable it. This should be updated when the translator is updated later on.

okankoAMZ (approved):

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM golang:1.20 as builder
+FROM golang:1.23 as builder
```
musa-asad, line 2: Should we be changing this?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/aws/amazon-cloudwatch-agent-operator
 
-go 1.21
+go 1.21.0
```
musa-asad, line 3: Should we be changing this?

go.mod
```diff
@@ -29,9 +29,10 @@ require (
 	github.com/prometheus/prometheus v0.48.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.9.0
+	go.opentelemetry.io/collector/confmap v0.102.1
```
musa-asad, line 32: Seems that v0.101.0 can be used instead to prevent changing the Go version.

okankoAMZ (approved):

PR #264, opened 2024-10-28
Setup cert-watcher for TA server cert

sky333999:
*Description of changes:* With the current setup, when the TA server certs are rotated, the TA pod is required to be restarted for it to pick up the updated certs - thus requiring explicit intervention.
With this change, we introduce a cert watcher that runs a background watch on the cert files and refreshes anytime a change is detected, thus still preserving connectivity without having to restart the TA pod.

*Testing*:
* Deployed the updated TA image to a dev cluster and executed the following sequence of events:
1. Helm install
2. Confirm collector is able to talk to TA
3. Helm upgrade with no change to simulate a rotation of certs
4. Restart collector pod to pick up new CA bundle on client i.e. collector (but not restart TA intentionally)
5. Confirm collector is still able to talk to TA
* The sequence above would fail prior to this change and passes with the change. Additionally, we see the following log line printed in TA pod logs:
```
{"level":"info","ts":"2024-10-28T23:26:44Z","logger":"controller-runtime.certwatcher","msg":"Updated current TLS certificate"}
```

### DIFF
--- a/cmd/amazon-cloudwatch-agent-target-allocator/config/config.go
+++ b/cmd/amazon-cloudwatch-agent-target-allocator/config/config.go
@@ -4,12 +4,14 @@
 package config
 
 import (
+	"context"
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
 	"fmt"
 	"io/fs"
 	"os"
+	"sigs.k8s.io/controller-runtime/pkg/certwatcher"
 	"time"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


cmd/amazon-cloudwatch-agent-target-allocator/config/config.go
```diff
@@ -207,26 +209,34 @@ func Load() (*Config, string, error) {
...
-	cert, err := tls.LoadX509KeyPair(c.TLSCertFilePath, c.TLSKeyFilePath)
-	if err != nil {
-		return nil, err
-	}
+func (c HTTPSServerConfig) NewTLSConfig(ctx context.Context) (*tls.Config, error) {
 	tlsConfig := &tls.Config{
-		Certificates: []tls.Certificate{cert},
-		ClientAuth:   tls.NoClientCert,
-		MinVersion:   tls.VersionTLS12,
+		MinVersion: tls.VersionTLS12,
```
okankoAMZ, line 221: even though it is not directly related to this PR , can we enforce TLS 1.3?

cmd/amazon-cloudwatch-agent-target-allocator/config/config.go
```diff
@@ -207,26 +209,34 @@ func Load() (*Config, string, error) {
...
+	}
+
+	if tlsConfig.GetCertificate == nil {
```
okankoAMZ, line 224: wouldn't this always be true?

okankoAMZ (commented):
Left few comments going to test it now

okankoAMZ (dismissed):
Tested seems to be working!

musa-asad (approved):

cmd/amazon-cloudwatch-agent-target-allocator/config/config.go
sky333999: yeah i can update that

musa-asad (approved):

PR #304, opened 2025-04-25
Splitting arm/amd tags for operator and target allocator

Paramadon:
# Split Multi-Architecture Image Builds for Better Security Scanning

## Problem
Currently, our operator and target allocator images are built as multi-architecture images, making it difficult to scan individual architecture images

## Solution
Split the build process into three parts:
1. Build individual architecture images
2. Create manifest lists
3. Clean up old images

### Flow Change
Before:
```mermaid
graph TD
    A[Build Multi-Arch Image] --> B[Push to ECR]
    B --> C[Security Scan]
    C --> D[Release]
```

After:
```mermaid
graph TD
    A1[Build AMD64 Image] --> B1[Push AMD64 to ECR]
    A2[Build ARM64 Image] --> B2[Push ARM64 to ECR]
    B1 --> C1[Security Scan AMD64]
    B2 --> C2[Security Scan ARM64]
    C1 --> D[Create Manifest List]
    C2 --> D
    D --> E[Release]
```

## Changes
1. Split image builds:
   - `staging-linux-amd64` for AMD64 architecture
   - `staging-linux-arm64` for ARM64 architecture
   - Manifest list pointing to both architectures

## Testing
- Built and verified AMD64 images
- Built and verified ARM64 images
- Created and verified manifest lists
- Tested image pulls for both architectures
- Verified security scanning works for individual architectures

## Verification
Screenshot 2025-04-25 at 3 20 52 PM


.github/workflows/build-and-upload.yml
```diff
@@ -157,20 +183,36 @@ jobs:
...
+          platforms: linux/amd64
+          tags: |
+            ${{ env.ECR_TARGET_ALLOCATOR_STAGING_REPO }}:${{ inputs.tag }}-linux-amd64
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          provenance: false
+
+      - name: Build Target Allocator ARM64 Image
         uses: docker/build-push-action@v4
-        if: steps.cached_binaries.outputs.cache-hit == false
```
lisguo, line 165: why was this line removed?

.github/workflows/build-and-upload.yml
```diff
@@ -87,17 +87,43 @@ jobs:
...
           push: true
-          platforms: linux/amd64,linux/arm64
+          platforms: linux/amd64
           tags: |
-            ${{ env.ECR_OPERATOR_STAGING_REPO }}:${{ inputs.tag }}
-            ${{ env.ECR_OPERATOR_STAGING_REPO }}:${{ inputs.target-sha }}
+            ${{ env.ECR_OPERATOR_STAGING_REPO }}:${{ inputs.tag }}-linux-amd64
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          build-args: |
```
lisguo, line 102: nit: the build args should be the same for arm and amd versions of the operator. Wondering if we can have a variable to store the build args or a workflow so that if we need to add Golang sdk or something, we update the build arges in 1 place

.github/workflows/build-and-upload.yml
Paramadon: So we removed caching in our agent repo because we only can have a 10gb limit and I think we are about to reach that. Guess we can have cleaners to take care of that, but I'll add this back.

.github/workflows/build-and-upload.yml
Paramadon: Good point! I added a env var and pointing to it for both arm and amd.

Paramadon: Good point! Added a seperate workflow for build-args so we can update args in one place.

.github/workflows/operator-build-args.yml
```diff
@@ -0,0 +1,30 @@
+name: 'Operator Build Args'
```
lisguo, line 1: instead of having a separate workflow just to obtain build args...I was thinking of a workflow to build the operator and it takes in an input for arch or tag.

I'd prefer the previous iteration over this

Paramadon: Ahh I see, I reverted it to the previous iteration.

.github/workflows/build-and-upload.yml
```diff
@@ -213,12 +258,64 @@ jobs:
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ env.AWS_ASSUME_ROLE }}
+          aws-region: ${{inputs.region}}
+
+      - name: Login ECR
+        uses: aws-actions/amazon-ecr-login@v2
+
+      - name: Create manifests
```
agarakan, line 300: Docker has a multi-tag feature thats also useful to prevent need for re-building a new image just to add the new tag. Not blocking comment

```
docker manifest create \
  ${{ env.ECR_TARGET_ALLOCATOR_STAGING_REPO }}:${{ inputs.tag }} \
  ${{ env.ECR_TARGET_ALLOCATOR_STAGING_REPO }}:${{ github.sha }} \
  --amend ${{ env.ECR_TARGET_ALLOCATOR_STAGING_REPO }}:${{ inputs.tag }}-linux-amd64 \
  --amend ${{ env.ECR_TARGET_ALLOCATOR_STAGING_REPO }}:${{ inputs.tag }}-linux-arm64

docker manifest push ${{ env.ECR_TARGET_ALLOCATOR_STAGING_REPO }}:${{ inputs.tag }}
docker manifest push ${{ env.ECR_TARGET_ALLOCATOR_STAGING_REPO }}:${{ github.sha }}
```

agarakan (approved):

.github/workflows/build-and-upload.yml
```diff
@@ -157,20 +184,38 @@ jobs:
         run: |
           go mod download
           go mod tidy
-          export GOARCH=arm64 && make targetallocator
-          export GOARCH=amd64 && make targetallocator 
+          mkdir -p cmd/amazon-cloudwatch-agent-target-allocator/bin
```
lisguo, line 187: why are we making a /bin directory here?

PR #320, opened 2025-06-10
Add support for UpdateStrategy when collector mode is Deployment

sky333999:
*Description of changes:* The CRD supports configuring UpdateStrategy when the collector mode is a DaemonSet, but not when its a Deployment. This extends support for configuring it for Deployments as well.


apis/v1alpha2/amazoncloudwatchagent_types.go
```diff
@@ -6,6 +6,7 @@
 package v1alpha2
```
mitali-salvi, line 6: outside the context of this PR, we should remove the v1alpha2 APIs since they are not being used.

mitali-salvi (commented):
Were we able to test this ?

PR #333, opened 2025-08-08
Ability to configure Collector's Service, Headless Service and Monitoring Service

sky333999:
### Description:

* Introduces new Collector CRD config options for configuring & managing the ClusterIP Service, Headless Service & Monitoring Service.
* For each of the Services, config to Enable/Disable the service and override the Name of the service.
* To maintain backwards compatibility, I considered the following two options and decided to go with the second.
  * Call the flag Disabled instead of Enabled such that it initializes to false when not set and the behavior stays the same.
  * Make the flags pointers instead of bools such that we can know whether the flag was explicitly provided i.e. nil implies retain old behavior.

### Description:
* Unit Testing: Added new unit tests
* Local Testing: Built image with changes, validated on dev cluster with helm chart to test creation & disabling of services.

Default installation:
image

Updated CR to disable ClusterIP service only:
image

Updated CR to disable all 3 services for cloudwatch-agent:
image

### DIFF
--- a/internal/manifests/collector/collector.go
+++ b/internal/manifests/collector/collector.go
@@ -52,7 +52,6 @@ func Build(params manifests.Params) ([]client.Object, error) {
 			if params.OtelCol.Spec.Service.IsEnabled() {
 				manifestFactories = append(manifestFactories, manifests.Factory(ServiceMonitor))
 			}
-			manifestFactories = append(manifestFactories, manifests.Factory(ServiceMonitor))
 		}
 	}
 	for _, factory := range manifestFactories {


config/rbac/role.yaml
```diff
@@ -72,6 +72,9 @@ rules:
   - cloudwatch.aws.amazon.com
   resources:
   - amazoncloudwatchagents
+  - dcgmexporters
+  - instrumentations
```
movence, line 76: are these updates expected?

internal/manifests/collector/collector.go
```diff
@@ -34,15 +34,24 @@ func Build(params manifests.Params) ([]client.Object, error) {
...
+		manifestFactories = append(manifestFactories, manifests.Factory(MonitoringService))
+	}
 	if params.OtelCol.Spec.Observability.Metrics.EnableMetrics && featuregate.PrometheusOperatorIsAvailable.IsEnabled() {
 		if params.OtelCol.Spec.Mode == v1alpha1.ModeSidecar {
 			manifestFactories = append(manifestFactories, manifests.Factory(PodMonitor))
 		} else {
+			if params.OtelCol.Spec.Service.IsEnabled() {
+				manifestFactories = append(manifestFactories, manifests.Factory(ServiceMonitor))
+			}
 			manifestFactories = append(manifestFactories, manifests.Factory(ServiceMonitor))
```
movence, line 55: should remove this?

movence (approved):

PR #339, opened 2025-09-15
Remove B3 propagator from default instrumentation

majanjua-amzn:
### Changes
B3 propagator is both no longer needed and a blocker as it currently overrides trace state and baggage which are needed for the new adaptive sampling feature.

Updated sleep time in integ test to ensure destroyed resources have had time to properly terminate. See the following run as an example of the test fetching the wrong pod: https://github.com/aws/amazon-cloudwatch-agent-operator/actions/runs/17743921110/job/50426338932#step:11:248

This filter is not working effectively: https://github.com/aws/amazon-cloudwatch-agent-operator/blob/main/integration-tests/manifests/cmd/validate_instrumentation_vars.go#L65

### Testing
Tested removal of B3 in Java-based sample environments by passing OTEL_PROPAGATORS environment variable with `tracecontext,baggage,xray` (in no specific order), ensuring no impact if B3 is removed

### DIFF
--- a/.github/workflows/operator-integration-test.yml
+++ b/.github/workflows/operator-integration-test.yml
@@ -74,7 +74,7 @@ jobs:
       - name: Test for default instrumentation resources for Java
         run: |
           kubectl apply -f integration-tests/java/sample-deployment-java.yaml
-          sleep 5
+          sleep 15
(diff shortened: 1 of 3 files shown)


lisguo:
Need to update unit test and e2e tests which verify the instrumentation CR

PR #360, opened 2026-03-10
fix: add restricted-compliant securityContext to Java init container

### LINKED ISSUE #361: Java init container missing securityContext breaks restricted Pod Security Standard
## What happened?

The Java auto-instrumentation init container (`opentelemetry-auto-instrumentation-java`) is created without any `securityContext`, causing pod creation to fail in namespaces enforcing `pod-security.kubernetes.io/enforce: restricted`.

## Error

```
Error creating: pods "app-789564bdf9-c6wm4" is forbidden: violates PodSecurity
"restricted:latest": allowPrivilegeEscalation != false (container
"opentelemetry-auto-instrumentation-java" must set
securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container
"opentelemetry-auto-instrumentation-java" must set
securityContext.capabilities.drop=["ALL"]), seccompProfile (pod or container
"opentelemetry-auto-instrumentation-java" must set
securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

## Root Cause

In `pkg/instrumentation/sdk.go`, `setInitContainerSecurityContext` is commented out for Java due to upstream opentelemetry-operator#2272 (runAsNonRoot conflict with root-based Java agent image). However, this leaves the init container with no `securityContext` at all, violating the restricted Pod Security Standard.

Other languages (NodeJS, Python, DotNet, Apache) are not affected — they all call `setInitContainerSecurityContext`.

## Expected behavior

The Java init container should receive a minimal restricted-compliant `securityContext`:
- `allowPrivilegeEscalation: false`
- `capabilities.drop: ["ALL"]`
- `seccompProfile.type: RuntimeDefault`

Without setting `runAsNonRoot` or `runAsUser` (to avoid the #2272 conflict).

## Environment

- EKS add-on: amazon-cloudwatch-observability
- Namespace PSS: `restricted:latest`
- Deployment: Pod-level `runAsUser: 1001`, `runAsNonRoot: true`, full restricted `securityContext` on app container

## Reproduction

1. Create namespace with `pod-security.kubernetes.io/enforce: restricted`
2. Install amazon-cloudwatch-observability EKS add-on
3. Deploy a Java app with `instrumentation.opentelemetry.io/inject-java: "true"` and restricted-compliant `securityContext`
4. Observe `FailedCreate` event on the ReplicaSet

## Fix

Submitted in #360.

ab0utbla-k:
Fixes #361

## Problem

The Java auto-instrumentation init container (`opentelemetry-auto-instrumentation-java`) is created without any `securityContext`, causing pod creation to fail in namespaces enforcing `pod-security.kubernetes.io/enforce: restricted`.

```
Error creating: pods "app-789564bdf9-c6wm4" is forbidden: violates PodSecurity
"restricted:latest": allowPrivilegeEscalation != false (container
"opentelemetry-auto-instrumentation-java" must set
securityContext.allowPrivilegeEscalation=false), unrestricted capabilities (container
"opentelemetry-auto-instrumentation-java" must set
securityContext.capabilities.drop=["ALL"]), seccompProfile (pod or container
"opentelemetry-auto-instrumentation-java" must set
securityContext.seccompProfile.type to "RuntimeDefault" or "Localhost")
```

Other languages (NodeJS, Python, DotNet, Apache) are not affected, because they all call `setInitContainerSecurityContext`, which copies the app container's security context onto the init container.

## Root Cause

In `pkg/instrumentation/sdk.go`, `setInitContainerSecurityContext` was commented out for Java to avoid a `runAsNonRoot` conflict with the root-based Java agent image (opentelemetry-operator#2272). However, this left the init container with no `securityContext` at all, violating the restricted Pod Security Standard.

## Fix

Added `setJavaInitContainerSecurityContext`, which derives the init container's context from the instrumented container instead of hardcoding one:

1. Deep-copy the instrumented container's `securityContext`, so user-specified fields (`readOnlyRootFilesystem`, `seLinuxOptions`, `runAsGroup`, `capabilities.add`, and so on) are preserved and behavior stays consistent with the other languages.
2. Strip `runAsNonRoot` and `runAsUser`, which are the fields that conflict with the root-based Java agent image (#2272).
3. Default the fields required by the restricted Pod Security Standard (`allowPrivilegeEscalation: false`, `capabilities.drop: ["ALL"]`, `seccompProfile.type: RuntimeDefault`), **only when the instrumented container leaves them unset**, so user values always win and a container with no `securityContext` at all still gets a compliant init container.

`setInitContainerSecurityContext` now deep-copies as well. It previously assigned the *same pointer* to both the init container and the instrumented container; stripping fields from a shared pointer would silently mutate the instrumented container. No behavior change today, since nothing mutated it before.

## Reproduction

1. Create namespace with `pod-security.kubernetes.io/enforce: restricted`
2. Install amazon-cloudwatch-observability EKS add-on
3. Deploy a Java app with `instrumentation.opentelemetry.io/inject-java: "true"` and a restricted-compliant `securityContext`
4. Observe `FailedCreate` event on the ReplicaSet

## Testing

- New `TestInjectJavaSecurityContext` covering three cases: no `securityContext` on the instrumented container (minimal restricted context applied), a full restricted context (`runAsNonRoot`/`runAsUser` dropped, everything else preserved), and a partial context (missing restricted fields defaulted). Each case also asserts the instrumented container's `securityContext` is not modified.
- Updated existing test cases that assert on the Java init container's expected pod spec (3 in `sdk_test.go`, 4 in `podmutator_test.go`) to include the resulting `securityContext`.
- `go build ./...`, `go vet ./...` and `go test ./pkg/instrumentation/...` pass.

### DIFF
--- a/pkg/instrumentation/sdk.go
+++ b/pkg/instrumentation/sdk.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/aws/amazon-cloudwatch-agent-operator/apis/v1alpha1"
@@ -99,9 +100,11 @@ func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations
 			} else {
 				pod = i.injectCommonEnvVar(otelinst, pod, index)
 				pod = i.injectCommonSDKConfig(ctx, otelinst, ns, pod, index, index)
-				//disable setting security context in init container due to issue with runAsNonRoot conflict
-				//https://github.com/open-telemetry/opentelemetry-operator/issues/2272
-				//pod = i.setInitContainerSecurityContext(pod, pod.Spec.Containers[index].SecurityContext, javaInitContainerName)
+
+				// Set a minimal restricted-compliant securityContext without runAsNonRoot/runAsUser
+				// to avoid the runAsNonRoot conflict (https://github.com/open-telemetry/opentelemetry-operator/issues/2272)
+				// while still satisfying the restricted Pod Security Standard.
+				pod = i.setInitContainerRestrictedSecurityContext(pod, javaInitContainerName)
 			}
 		}
 	}
@@ -297,6 +300,23 @@ func (i *sdkInjector) setInitContainerSecurityContext(pod corev1.Pod, securityCo
 		}
 	}
 
+	return pod
+}
+
+func (i *sdkInjector) setInitContainerRestrictedSecurityContext(pod corev1.Pod, instrInitContainerName string) corev1.Pod {
+	for idx, initContainer := range pod.Spec.InitContainers {
+		if initContainer.Name == instrInitContainerName {
+			pod.Spec.InitContainers[idx].SecurityContext = &corev1.SecurityContext{
+				AllowPrivilegeEscalation: ptr.To(false),
+				Capabilities: &corev1.Capabilities{
+					Drop: []corev1.Capability{"ALL"},
+				},
+				SeccompProfile: &corev1.SeccompProfile{
+					Type: corev1.SeccompProfileTypeRuntimeDefault,
+				},
+			}
+		}
+	}
 	return pod
 }
 
--- a/pkg/instrumentation/sdk.go
+++ b/pkg/instrumentation/sdk.go
@@ -99,9 +99,8 @@ func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations
 			} else {
 				pod = i.injectCommonEnvVar(otelinst, pod, index)
 				pod = i.injectCommonSDKConfig(ctx, otelinst, ns, pod, index, index)
-				//disable setting security context in init container due to issue with runAsNonRoot conflict
-				//https://github.com/open-telemetry/opentelemetry-operator/issues/2272
-				//pod = i.setInitContainerSecurityContext(pod, pod.Spec.Containers[index].SecurityContext, javaInitContainerName)
+
+				pod = i.setJavaInitContainerSecurityContext(pod, pod.Spec.Containers[index].SecurityContext, javaInitContainerName)
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


mrowken:
Hi, we need this feature, kindly assign reviewer please.

pkg/instrumentation/sdk.go
```diff
@@ -99,9 +100,11 @@ func (i *sdkInjector) inject(ctx context.Context, insts languageInstrumentations
...
 				pod = i.injectCommonEnvVar(otelinst, pod, index)
 				pod = i.injectCommonSDKConfig(ctx, otelinst, ns, pod, index, index)
-				//disable setting security context in init container due to issue with runAsNonRoot conflict
-				//https://github.com/open-telemetry/opentelemetry-operator/issues/2272
-				//pod = i.setInitContainerSecurityContext(pod, pod.Spec.Containers[index].SecurityContext, javaInitContainerName)
+
+				// Set a minimal restricted-compliant securityContext without runAsNonRoot/runAsUser
+				// to avoid the runAsNonRoot conflict (https://github.com/open-telemetry/opentelemetry-operator/issues/2272)
+				// while still satisfying the restricted Pod Security Standard.
+				pod = i.setInitContainerRestrictedSecurityContext(pod, javaInitContainerName)
```
mitali-salvi: seems like this security context is missing in NGINX instrumentation as well - https://github.com/ab0utbla-k/amazon-cloudwatch-agent-operator/blob/3969c9f684e19dbd5fb5be421c6d72183b95036e/pkg/instrumentation/sdk.go#L231 mind adding there as well ?

mitali-salvi:
The other languages (Node.js, Python, .NET, Apache) use setInitContainerSecurityContext which copies the app container's full securityContext onto the init container. This PR introduces a separate
setInitContainerRestrictedSecurityContext that applies a hardcoded minimal context instead.

Have you considered modifying the existing setInitContainerSecurityContext to copy the app container's securityContext but strip runAsNonRoot and runAsUser for the Java case? That would preserve any
additional user-specified fields (e.g. readOnlyRootFilesystem, seLinuxOptions, custom capabilities.add) while still avoiding the #2272 conflict — and keep the behavior consistent across all languages.

Pawelgasieniec (approved):

Pawelgasieniec (approved):

pkg/instrumentation/sdk.go
ab0utbla-k: Thanks for the catch, but I think NGINX is already covered. It just sets the context inline in `nginx.go` rather than through the `sdk.go` helper, which is why it looks missing at the line linked:

- `otel-agent-attach-nginx`: [`nginx.go#L169`](https://github.com/aws/amazon-cloudwatch-agent-operator/blob/9a8bdb6587b8678118dc4fec36da32c903220b0b/pkg/instrumentation/nginx.go#L169) sets `SecurityContext: pod.Spec.Containers[index].SecurityContext`
- `otel-agent-source-container-clone`: created via `container.DeepCopy()` at [`nginx.go#L78`](https://github.com/aws/amazon-cloudwatch-agent-operator/blob/9a8bdb6587b8678118dc4fec36da32c903220b0b/pkg/instrumentation/nginx.go#L78), so it inherits the instrumented container's context

Happy to move both to the shared helper for consistency if you'd prefer, but that felt out of scope for this fix. Java was the only language ending up with no context at all.

ab0utbla-k:
@mitali-salvi good call, done, the PR now does exactly that.

`setJavaInitContainerSecurityContext` copies the instrumented container's `securityContext`, strips `runAsNonRoot`/`runAsUser`, and keeps everything else (`readOnlyRootFilesystem`, `seLinuxOptions`, `capabilities.add`, and so on).

Two things worth flagging:

1. **The copy has to be deep.** `setInitContainerSecurityContext` assigned the *same pointer* to the init container and the instrumented container. Stripping `runAsNonRoot` on that shared pointer would also remove it from the instrumented container. I changed the shared helper to `DeepCopy()`. No behavior change today since nothing mutated it, but it makes the strip safe. There is a test asserting the instrumented container is untouched.
2. **Nil fallback.** If the instrumented container has no `securityContext`, a plain copy reproduces the original bug. The restricted fields (`allowPrivilegeEscalation`, `capabilities.drop`, `seccompProfile`) are therefore defaulted only when unset, so user values always win.

Added `TestInjectJavaSecurityContext` covering all three paths, and rebased on latest `main`.

PR #380, opened 2026-05-05
Add language detection to auto-monitor to inject only the relevant SDK

Miqueasher:
## Summary

When `monitorAllServices: true`, auto-monitor currently injects all 4 language SDK init containers (Java, Python, Node.js, .NET) into every pod regardless of runtime, causing liveness/readiness probe failures, restart loops, and deployment instability.

This PR adds a registry-based language detector that inspects container image config (ENV, CMD, ENTRYPOINT) via google/go-containerregistry without pulling layers (~100-500ms, 5s timeout)
Falls back gracefully through image name patterns → pod-spec env vars → pod-spec commands → all languages (current behavior), ensuring zero regression

## Detection Layers
1. Registry image config — fetch ENV/CMD/ENTRYPOINT from manifest (public registries + ECR with IRSA)
2. Image name patterns — match language keywords in image reference
3. Pod-spec env vars — check for JAVA_HOME, PYTHONPATH, NODE_ENV, DOTNET_ROOT, etc.
4. Pod-spec commands — match runtime binaries (java, python, node, dotnet)
5. Fallback — all configured languages

## Dependencies added
Package | Version | Purpose
github.com/google/go-containerregistry | v0.20.0 | Fetch image config from registry without pulling layers
github.com/aws/aws-sdk-go (existing) | v1.45.25 | ECR auth via custom keychain

## Test Plan
* Unit tests: 92/92 passing (65 new + 27 existing, zero regressions)
* Live EKS validation (us-east-1): Custom operator build deployed with monitorAllServices=true, no manual annotations
* Java, Python, Node.js, .NET public images → 1 init container each (detected via registry config)
* Alpine, Nginx, Busybox → 4 init containers (correct fallback, no false positives)
* Private ECR image with language keyword → detected via image name pattern
* Pod-spec JAVA_HOME on Alpine → detected via env var fallback
* Multi-container (Python + Nginx sidecar) → correctly detected Python only
* Failure mode: When all detection fails, behavior is identical to current production
*  Operator image digest confirmed matching between running pod and ECR (sha256:5f74a89f76b3...), zero pod restarts observed


pkg/instrumentation/auto/language_detector.go
```diff
@@ -0,0 +1,353 @@
...
+
+	return ""
+}
+
+// fetchImageConfig retrieves the image config (ENV, CMD, ENTRYPOINT, Labels) from the
+// registry. Only fetches the manifest and config blob — no layer data is downloaded.
+func (d *languageDetector) fetchImageConfig(imageRef string) *v1.Config {
+	ref, err := name.ParseReference(imageRef)
+	if err != nil {
+		d.logger.V(2).Info("could not parse image reference", "image", imageRef, "error", err)
```
mxiamxia: should it be Error logs instead of Info? the same applies the other places

pkg/instrumentation/auto/monitor.go
```diff
@@ -450,16 +452,32 @@ func getTemplateSpecLabels(obj metav1.Object) labels.Set {
...
 	languagesToAnnotate := m.config.CustomSelector.LanguagesOf(obj, false)
 	if m.isWorkloadAutoMonitored(obj) {
-		for l := range m.config.Languages {
-			languagesToAnnotate[l] = nil
+		// Attempt to detect the language from container spec before falling back to all languages
+		detected := m.detectLanguagesFromWorkload(obj)
+		if len(detected) > 0 {
+			m.logger.V(1).Info("auto-monitor detected language(s) from container spec",
+				"objName", obj.GetName(), "detected", detected)
+			for l := range detected {
```
mxiamxia, line 470: we should only expect one language be returned from App Container. I think the case you're supporting includes all sidecar container(s), can we distinguish App vs. Sidecar container?

pkg/instrumentation/auto/language_detector.go
```diff
@@ -0,0 +1,353 @@
...
+
+// detectLanguages inspects all containers in a pod template and returns the set of
+// detected languages. Returns an empty set if no language can be confidently determined.
+func (d *languageDetector) detectLanguages(podSpec *corev1.PodTemplateSpec) instrumentation.TypeSet {
+	detected := make(instrumentation.TypeSet)
+
+	for _, container := range podSpec.Spec.Containers {
+		if lang := d.detectContainer(container); lang != "" {
+			d.logger.V(2).Info("detected language from container",
+				"container", container.Name, "image", container.Image, "language", lang)
```
mxiamxia, line 35: I think we're printing lots of logs which are redundent here and there, we should remove those redundant and turn most of them into debug logs

pkg/instrumentation/auto/language_detector.go
```diff
@@ -0,0 +1,353 @@
...
+
+// detectContainer fetches the image config from the registry and inspects it.
+// Falls back to pod-spec-only detection if the registry fetch fails.
+func (d *languageDetector) detectContainer(container corev1.Container) instrumentation.Type {
+	// Try fetching real image config from registry
+	if cfg := d.fetchImageConfig(container.Image); cfg != nil {
+		if lang := d.detectFromConfig(cfg); lang != "" {
+			return lang
+		}
+	}
```
mxiamxia: detectFromConfig actually contains the remote call to registries which is costly. Can we move deprioritize this and make it as the last fallback. Prioritize others which is cheap checks, if the language can be decided from there we don't need to make remote call

pkg/instrumentation/auto/language_detector.go
```diff
@@ -0,0 +1,353 @@
...
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), d.timeout)
+	defer cancel()
+
+	desc, err := remote.Get(ref,
+		remote.WithAuthFromKeychain(d.keychain),
+		remote.WithContext(ctx),
+	)
```
mxiamxia: I think we need to cache this for the same kind of pod. otherwise for large cluster with thousands of pod, it will either be throttled or make some performance impact.

pkg/instrumentation/auto/language_detector.go
```diff
@@ -0,0 +1,353 @@
...
+// ecrKeychain uses the AWS SDK default credential chain (instance profile, IRSA, env vars)
+// to authenticate with Amazon ECR. This is the same credential chain customers configure
+// via aws configure / IAM roles when setting up EKS.
+type ecrKeychain struct {
+	logger logr.Logger
+}
+
+func newECRKeychain(logger logr.Logger) *ecrKeychain {
+	return &ecrKeychain{logger: logger}
+}
```
mxiamxia: what if the cluster role doesn't have ECR permissions, will the detection logic be failed gracefully without impact service deployment, and then fall back to other default detections or eventually inject all 4 SDKs as today. we need to handle errors gracefully.

Miqueasher: Yes, if ECR permissions are missing, detection fails gracefully and falls back to current behavior (all 4 SDKs injected). No deployment is impacted.

The detection has multiple fallback layers. So if the cluster role lacks ECR permissions:

Layer 1: ECR auth fails -> no error raised, proceeds with unauthenticated request

Layer 2: Unauthenticated registry call returns 401/403 -> caught, returns nil

Layer 3: With no image config, falls to local detection using data already in the pod spec

Layer 4: If local detection also can't determine the language -> injects all 4 SDKs, identical to today's behavior

pkg/instrumentation/auto/language_detector.go
Miqueasher: All log statements in the detection logic are already at V(2) (only visible when operator is started with --zap-log-level=2 or higher). None are at default info level.

Please see the logr. documentation: https://github.com/go-logr/logr

The V(1) logs were placed in order to maintain the current pattern of logging seen in monitor.go:

V(1): operational milestones/outcomes
V(2): granular internal states

I'm not seeing any redundancy, but I'd be happy to consolidate if I could have them pointed out!

pkg/instrumentation/auto/language_detector.go
Miqueasher: Good call, I'll reorder to try the cheap local checks first and only make the registry call if they're inconclusive!

pkg/instrumentation/auto/language_detector.go
Miqueasher: I made these intentionally Info-level since they represent expected operational conditions. An Error would signal that action is needed or something is broken, but in these the system is working as designed by falling through to the next detection layer.

pkg/instrumentation/auto/language_detector.go
Miqueasher: Agreed — we'll add an image-level cache so that the same image reference is only looked up once from the registry. If 200 workloads use the same image, the registry call fires once and subsequent lookups return the cached result.

pkg/instrumentation/auto/monitor.go
Miqueasher: Currently there's no reliable Kubernetes-native way to distinguish app containers from sidecars in the containers[] array that I have found. We could maybe implement a skip-list of sorts where we store a list of known sidecars that get excluded from detection.

Even that wouldn't fully solve the issue though.

mxiamxia (commented):
The new revision looks good to me. Pls resolve the conflicts and I'll approve CR from my side.

v-iw:
Hi @Miqueasher — this PR has conflicts with `main` and I don't have permissions to push to your fork, so I've rebased your commits onto current `main` and opened #409 to carry it forward. Your two commits are preserved as-is with you as the author.

The three source files in #409 are byte-identical to this PR at `acf4dc8`. The only thing I dropped is the `go.mod` / `go.sum` change: it bumped `opencontainers/image-spec` to `v1.1.0-rc3`, which was left over from the registry-inspection approach that your second commit ("removing network call") already removed. `main` is now on `v1.1.1` and nothing in the final code imports `go-containerregistry`, so it was both unused and a downgrade — and it was the entire source of the conflict.

Could you take a look at #409 and confirm the rebase didn't drop anything you intended to keep? Happy to close this one out once you've verified.

PR #390, opened 2026-06-30
Allocation Strategy and TA resilience

wenegiemepraise:
*Issue #, if available:*

*Description of changes:*
Adds per-node target allocation to the Target Allocator (TA) and makes the TA resilient to ServiceMonitor/PodMonitor CRD install ordering, so the SM/PM scraping path works zero-step and node-locally on a DaemonSet agent deployment.

The TA previously built SM/PM informers once at startup and blocked on cache sync; a missing CRD failed the sync and required a restart to pick the CRD up later.
Now the TA watches CustomResourceDefinition objects and lazily starts each monitor informer when its CRD appears 


Aakash-Dantre:
lgtm!

PR #386, opened 2026-06-16
Fix Target Allocator startup crashes, PrometheusCR watcher, and Prometheus config pod restart

musa-asad:
## Summary

Multiple defects prevent the Target Allocator (TA) from working correctly in both Helm and EKS Add-On deployments:

1. **TA CrashLoopBackOff:** The TA binary never registered the `--enable-prometheus-cr-watcher` flag the operator passes, so `pflag.ExitOnError` calls `os.Exit(2)` immediately.
2. **Three latent defects** (unreachable until fix §1): synthetic Prometheus empty-namespace panic, empty `evaluation_interval` parse failure, nil SD metrics → 0 targets discovered.
3. **Prometheus config changes don't restart agent pods:** Only `spec.config` changes triggered restarts — `spec.prometheus` changes were ignored.
4. **`scrape_protocols` regression risk:** No test guarded that the Prometheus dependency defaults `scrape_protocols` on config load.

### TA binary (commit 1):
- `config/flags.go` — register `enable-prometheus-cr-watcher` Bool flag + getter
- `config/config.go` — OR the flag into `Config.PrometheusCR.Enabled`
- `watcher/promOperator.go` — set `ObjectMeta.Namespace` from `OTELCOL_NAMESPACE` env; set `EvaluationInterval` to `ScrapeInterval`
- `main.go` — initialize `sdMetrics` with `discovery.CreateAndRegisterSDMetrics()`

### Pod restart (commit 3):
- `internal/manifests/collector/annotations.go` — `configHashInput()` appends serialized `Spec.Prometheus` to the hash, gated by `Prometheus.IsEmpty()`

### Regression test (commit 2):
- `config/scrape_protocols_regression_test.go` — `TestScrapeProtocolsDefaultedOnLoad`

### A/B Comparison — Upstream OTel vs CWA Helm (fixed) vs CWA Add-On (fixed)

All 3 environments deployed on the same EKS cluster, same OTEL version, same scrape targets. The fixed images were deployed on both the Helm chart path AND the EKS managed Add-On path (add-on installed via `aws eks create-addon`, then TA/operator deployments patched to the fixed images).

| Check | upstream-otel | cwa-helm (fixed) | cwa-addon (fixed) |
|-------|---------------|-------------------|-------------------|
| TA pod Running | Running 14h, 0 restarts | Running 21m, 0 restarts | Running 11h, 0 restarts |
| /livez | 200 OK | 200 (mTLS) | 200 (mTLS) |
[...]

PR #394, opened 2026-07-06
fix(target-allocator): tolerate missing ServiceMonitor/PodMonitor CRDs

wenegiemepraise:
## Summary
  Make the Target Allocator (TA) start and run healthily whether or not the community
  `monitoring.coreos.com` ServiceMonitor/PodMonitor CRDs are installed, and begin/stop
  watching each type automatically as its CRD appears/disappears — with no restart, ever.
  Built on top of this PR: https://github.com/aws/amazon-cloudwatch-agent-operator/pull/386

 ## Problem
  The TA *consumes* the SM/PM CRDs but doesn't require them to exist. Today
  `Watch()` builds the ServiceMonitor **and** PodMonitor informers unconditionally at
  startup and blocks on `WaitForNamedCacheSync`. If a CRD is absent, the apiserver
  rejects LIST/WATCH for that GVR, the informer never syncs, and `Watch()` returns
  `failed to sync cache`. Because that setup runs once, installing the CRD later
  requires a **TA restart**.

  ## Fix
  Make the watcher event-driven and per-type instead of one-shot and all-or-nothing:
  - `crdExists` — check whether a given CRD is installed before building its informer.
  - `startMonitorInformer` / `stopMonitorInformer` — build/tear down a single monitor
    informer on demand (idempotent); stopping drops its targets and signals a reload.
  - `watchCRDs` — an informer over `CustomResourceDefinition` objects; on Add starts the
    corresponding monitor informer, on Delete stops it.
  - `Watch()` rewritten — start informers only for CRDs present at startup; absent ones
    are skipped (TA stays healthy) and started later by the CRD watch. Never returns
    `failed to sync cache` due to a missing CRD.

  ## Testing
  **1. Unit (this PR):**
  - `TestWatchStartsHealthyWithoutCRDs` — no CRDs → Watch healthy, no error, no jobs
  - `TestWatchStartsInformerWhenCRDCreated` — CRD created at runtime → informer starts, no restart
  - `TestWatchPerTypeIndependence` — SM present, PM absent → only SM informer starts
  - `TestStopMonitorInformerDropsType` — CRD removed → informer stopped, targets dropped, reload signalled
  - `TestCRDObjectName` — CRD-name extraction incl. delete tombstone
  
  **2. Before/after reproduction (A/B):** same condition (CRDs absent) —
  pre-fix `Watch()` returns `failed to sync cache` (output above); post-fix
  `TestWatchStartsHealthyWithoutCRDs` passes (healthy) and `TestWatchStartsInformerWhenCRDCreated`
  self-heals when the CRD appears.


cmd/amazon-cloudwatch-agent-target-allocator/watcher/promOperator.go
```diff
@@ -105,65 +155,198 @@ func getSelector(s map[string]string) labels.Selector {
...
+		return err
 	}
 
-	return map[string]*informers.ForResource{
-		monitoringv1.ServiceMonitorName: serviceMonitorInformers,
-		monitoringv1.PodMonitorName:     podMonitorInformers,
-	}, nil
+	stopCh := make(chan struct{})
+	informer.Start(stopCh)
+	if ok := cache.WaitForNamedCacheSync(resourceName, stopCh, informer.HasSynced); !ok {
```
musa-asad: `startMonitorInformer` holds `informersMtx` across `Start` and `WaitForNamedCacheSync`, and `stopCh` isn't recorded in `informerStopChannels` yet, so a slow apiserver could block the other start and stop paths and keep `Close()` from interrupting the sync. Could we run the sync outside the lock and select on `w.stopChannel`?

wenegiemepraise: You're right i didnt catch this. I reproduced and fixed the issue. The sync ran under informersMtx and only watched a local stop channel, so a slow/rejected LIST hung Close(). Now Start/WaitForNamedCacheSync run outside the lock and select on w.stopChannel, and Close() signals shutdown before draining. Added a test that blocks the LIST and asserts Close() returns (hangs before, passes after, race-clean).

cmd/amazon-cloudwatch-agent-target-allocator/watcher/promOperator.go
```diff
@@ -105,64 +155,235 @@ func getSelector(s map[string]string) labels.Selector {
...
-				}
-			},
-		})
+// watchCRDs starts a CustomResourceDefinition informer that starts/stops the
+// per-type monitoring informers as the ServiceMonitor/PodMonitor CRDs are
+// created or deleted at runtime — no TA restart required. The informer replays
+// existing CRDs on its initial sync, so CRDs present at startup are handled here
+// too (startMonitorInformer is idempotent).
+func (w *PrometheusCRWatcher) watchCRDs(notifyEvents chan struct{}) {
+	factory := apiextensionsinformers.NewSharedInformerFactory(w.crdClient, allocatorconfig.DefaultResyncTime)
```
musa-asad: `watchCRDs` informs on every CRD and caches each full object with its OpenAPI schema, even though only the two names in `crdNameToResource` matter, so memory could balloon on clusters with lots of CRDs. Could a `PartialObjectMetadata` informer or a `metadata.name` selector on those two names work instead?

cmd/amazon-cloudwatch-agent-target-allocator/watcher/promOperator.go
```diff
@@ -105,64 +155,235 @@ func getSelector(s map[string]string) labels.Selector {
...
+// TA starts and stays healthy regardless of CRD install ordering, and begins
+// (or stops) watching each type automatically as its CRD is created or deleted.
+func (w *PrometheusCRWatcher) Watch(upstreamEvents chan Event, upstreamErrors chan error) error {
+	// this channel needs to be buffered because notifications are asynchronous and neither producers nor consumers wait
+	notifyEvents := make(chan struct{}, 1)
+
+	// Start informers for CRDs that already exist. Absent CRDs are simply
+	// skipped here; the CRD watch below starts them if/when they appear.
+	for crdName, resourceName := range crdNameToResource {
+		exists, err := w.crdExists(context.Background(), crdName)
```
musa-asad: `crdExists` in the `Watch` startup loop calls `Get` with `context.Background()` and no deadline, so a wedged apiserver could block the loop forever. Could we pass a context with a timeout?

cmd/amazon-cloudwatch-agent-target-allocator/watcher/promOperator.go
```diff
@@ -204,32 +425,56 @@ func (w *PrometheusCRWatcher) rateLimitedEventSender(upstreamEvents chan Event,
 }
 
 func (w *PrometheusCRWatcher) Close() error {
+	// Signal shutdown first so any in-flight startMonitorInformer aborts its cache
+	// sync and will not register a new informer after this point.
 	close(w.stopChannel)
```
musa-asad: `Close()` will panic if it's ever called twice since it closes `w.stopChannel` unconditionally. Only called once today, but a `sync.Once` would be cheap insurance.

cmd/amazon-cloudwatch-agent-target-allocator/watcher/promOperator.go
```diff
@@ -105,64 +155,235 @@ func getSelector(s map[string]string) labels.Selector {
...
-		})
+// watchCRDs starts a CustomResourceDefinition informer that starts/stops the
+// per-type monitoring informers as the ServiceMonitor/PodMonitor CRDs are
+// created or deleted at runtime — no TA restart required. The informer replays
+// existing CRDs on its initial sync, so CRDs present at startup are handled here
+// too (startMonitorInformer is idempotent).
+func (w *PrometheusCRWatcher) watchCRDs(notifyEvents chan struct{}) {
+	factory := apiextensionsinformers.NewSharedInformerFactory(w.crdClient, allocatorconfig.DefaultResyncTime)
+	crdInformer := factory.Apiextensions().V1().CustomResourceDefinitions().Informer()
+	_, _ = crdInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
musa-asad: The `AddEventHandler` error is dropped here and in `startMonitorInformer`, so a failed registration would silently wire up no handler. Could we at least log it so a dead watch is visible?

cmd/amazon-cloudwatch-agent-target-allocator/watcher/promOperator_test.go
```diff
@@ -445,3 +471,139 @@ func sanitizeScrapeConfigsForTest(scs []*promconfig.ScrapeConfig) {
...
+		return runningInformers(w)[monitoringv1.ServiceMonitorName]
+	}, 5*time.Second, 10*time.Millisecond)
+}
+
+// TestStopMonitorInformerDropsType verifies the stop path used when a CRD is
+// removed at runtime: the informer is stopped, dropped from the active set (so
+// LoadConfig no longer emits its targets), and a reload is signalled. This is
+// the logic invoked by the CRD watch's DeleteFunc; it is exercised directly so
+// the assertion does not depend on fake-clientset delete-watch delivery.
+func TestStopMonitorInformerDropsType(t *testing.T) {
```
musa-asad, line 594: The Delete path from `crdObjectName` to `stopMonitorInformer` on the CRD informer's `DeleteFunc` is the one bit not covered end to end. Could we add a test that deletes the CRD via the fake client and asserts the informer stops, or confirm the E2E suite already hits it?

musa-asad (dismissed):

Aakash-Dantre:
Checklist posted for the PR author to move into the PR description under a `## PR Checklist` heading.

## PR Checklist

- [ ] Commits are squashed into a logical, reviewable set (one commit for a single change)
- [x] Commits and PR description comply with Amazon internal guidelines
- [x] `make` passes locally — `go build ./...` clean, `make impi` PASS, `make checklicense` PASS, unit tests PASS for all target-allocator and manifests packages
- [ ] All GitHub Actions checks on the PR are passing
- [ ] Integration test evidence: link to a passing run, or state N/A with the reason
- [x] New or updated integration test coverage: aws/amazon-cloudwatch-agent-test#720 (`ta_resilience_test.go`, `crd_bundling_test.go`)
- [x] New functionality has unit tests; bug fixes have a reproducing test
- [x] Config translation changes include updated golden files — N/A, no translator changes
- [ ] Breaking or customer-visible changes are called out in the PR description

> Note for the description: this change makes the Target Allocator require `get;list;watch` on
> `customresourcedefinitions`. Without the companion chart grant (aws-observability/helm-charts#327)
> the CRD check fails and ServiceMonitor/PodMonitor discovery silently yields no targets while the
> Target Allocator still reports healthy. Worth calling out as an operational prerequisite.

PR #398, opened 2026-07-09
[amazon-cloudwatch-agent-operator] feat(target-allocator): add per-node allocation strategy

wenegiemepraise:
## Summary
Add a **per-node** allocation strategy to the Target Allocator so each CloudWatch Agent
(DaemonSet, one per node) scrapes only the ServiceMonitor/PodMonitor targets on **its own
node** — eliminating cross-node / cross-AZ scrape traffic. Targets without a resolvable node
fall back to consistent-hashing, so nothing is silently dropped.

## Motivation
The fork registers only `consistent-hashing`, a pure hash of the target URL with no node
awareness — a pod on node A can be scraped by the agent on node B (inter-AZ data-transfer
cost + latency). This ports the upstream OpenTelemetry `per-node` strategy into the fork.

## Changes
- `allocation/per_node.go` — node-indexed allocator with consistent-hashing fallback and
  unassigned-target tracking
- `allocation/strategy.go` — register `per-node`
- `collector/collector.go` — capture `Collector.NodeName` from `pod.Spec.NodeName`
  (skip empty-NodeName pods; handle `watch.Modified`)
- CRD allocation-strategy enum (`consistent-hashing | per-node`)
- `internal/manifests/targetallocator/configmap.go` — emit `allocation_strategy` +
  `allocation_fallback_strategy`

## Dependencies
- **Depends on #386** (TA startup fixes). Until #386 merges, this PR's diff includes its
  commits. (Note: part of #386 — the SD-metrics registration — has already landed on `main`.)
- **Companion:** helm PR aws-observability/helm-charts#330 (chart per-node support +
  node-enrichment transforms). The chart default `per-node` requires this strategy registered
  in the TA build.

## Testing
- `allocation` unit tests (per-node placement, fallback, unassigned tracking)
- A/B: per-node vs consistent-hashing — per-node yields node-local assignment; flipping to
  consistent-hashing reproduces cross-node scrapes (validated live on EKS).


cmd/amazon-cloudwatch-agent-target-allocator/allocation/per_node.go
```diff
@@ -0,0 +1,396 @@
...
+	// fallbackHasher, when non-nil, is a consistent-hashing ring over all
+	// collectors used to place targets that cannot be matched to a node.
+	fallbackHasher *consistent.Consistent
+
+	log logr.Logger
+
+	filter Filter
+}
+
+func newPerNodeAllocator(log logr.Logger, opts ...AllocationOption) Allocator {
```
musa-asad, line 84: `newPerNodeAllocator` leaves `fallbackHasher` nil unless `SetFallbackStrategy` is called, so a hand written `per-node` config with no fallback keeps targets that have no node label but never scrapes them. Safe today because `configmap.go` emits `consistent-hashing`, but could we default it or warn when `per-node` has no fallback?

cmd/amazon-cloudwatch-agent-target-allocator/allocation/per_node.go
```diff
@@ -0,0 +1,396 @@
...
+	// Insert the new collectors.
+	for _, i := range diff.Additions() {
+		pn.collectors[i.Name] = NewCollector(i.Name, i.NodeName)
+		if pn.fallbackHasher != nil {
+			pn.fallbackHasher.Add(pn.collectors[i.Name])
+		}
+	}
+
+	// Rebuild the node index from the current collector set.
+	pn.collectorByNode = make(map[string]*Collector)
```
musa-asad, line 261: `collectorByNode[c.NodeName] = c` has no tie break, so two collectors on the same `NodeName` means last one wins and its targets flap. Can't happen at one pod per node, but a `maxSurge` DaemonSet rollout could hit it. Could we add a tie break (say the smaller pod name) or note the one pod per node assumption?

cmd/amazon-cloudwatch-agent-target-allocator/allocation/per_node.go
```diff
@@ -0,0 +1,396 @@
...
+}
+
+// SetCollectors sets the set of collectors with key=collectorName, value=Collector object.
+// This method is called when Collectors are added or removed.
+func (pn *perNodeAllocator) SetCollectors(collectors map[string]*Collector) {
+	timer := prometheus.NewTimer(TimeToAssign.WithLabelValues("SetCollectors", perNodeStrategyName))
+	defer timer.ObserveDuration()
+
+	CollectorsAllocatable.WithLabelValues(perNodeStrategyName).Set(float64(len(collectors)))
+	if len(collectors) == 0 {
```
musa-asad, line 364: `SetCollectors` returns early on an empty set without clearing its maps, so `GetTargetsForCollectorAndJob` keeps serving dead collectors until the next non empty call. `consistentHashingAllocator` does the same, so I read it as intentional parity, but if you want `per-node` strictly correct we could clear the maps at zero.

cmd/amazon-cloudwatch-agent-target-allocator/allocation/strategy.go
```diff
@@ -92,9 +120,12 @@ var _ consistent.Member = Collector{}
 
 // Collector Creates a struct that holds Collector information.
 // This struct will be parsed into endpoint with Collector and jobs info.
+// NodeName is the Kubernetes node the collector pod runs on; it is used by the
+// per-node allocation strategy to match targets to the collector on their node.
 // This struct can be extended with information like annotations and labels in the future.
 type Collector struct {
```
musa-asad, line 126: `Collector.Hash()` only returns `Name`, so a `Modified` event that changes `NodeName` on the same pod produces no diff and `collectorByNode` never rebuilds. It's safe because `spec.NodeName` is immutable and unscheduled pods are skipped, but that's subtle, so could we add a comment or hash `Name` plus `NodeName`?

cmd/amazon-cloudwatch-agent-target-allocator/main.go
```diff
@@ -73,7 +73,9 @@ func main() {
 	log := ctrl.Log.WithName("allocator")
 
 	allocatorPrehook = prehook.New(cfg.GetTargetsFilterStrategy(), log)
-	allocator, err = allocation.New(cfg.GetAllocationStrategy(), log, allocation.WithFilter(allocatorPrehook))
+	allocator, err = allocation.New(cfg.GetAllocationStrategy(), log,
```
musa-asad, line 76: Since this PR is what first registers `per-node`, an older allocator image without it hits `allocation.New` erroring and `os.Exit(1)`, so it just `CrashLoopBackOff`s instead of degrading. Could the companion chart pin the Target Allocator image at this commit or newer when it sets the strategy?

musa-asad (dismissed):

cmd/amazon-cloudwatch-agent-target-allocator/collector/collector.go
```diff
@@ -117,8 +120,15 @@ func runWatch(ctx context.Context, k *Client, c <-chan watch.Event, collectorMap
 			}
 
 			switch event.Type { //nolint:exhaustive
-			case watch.Added:
-				collectorMap[pod.Name] = allocation.NewCollector(pod.Name)
+			case watch.Added, watch.Modified:
```
movence, line 123: The List path above skips pods with a DeletionTimestamp but this watch branch does not, so a terminating pod keeps its node ownership until the Deleted event lands. Could we add the same check here?

cmd/amazon-cloudwatch-agent-target-allocator/allocation/per_node.go
```diff
@@ -0,0 +1,419 @@
...
+				reason = "no collector running on node " + nodeName
+			}
+			pn.log.V(1).Info("per-node: no node-local collector, used consistent-hashing fallback",
+				"target", strings.Join(tg.TargetURL, ","), "job", tg.JobName, "node", nodeName, "collector", fallbackCol.Name, "reason", reason)
+			return placedByFallback
+		}
+	}
+	if pn.fallbackHasher == nil && !pn.warnedNoFallback {
+		pn.warnedNoFallback = true
+		pn.log.Info("per-node: no fallback strategy configured; targets that cannot be matched to a " +
```
movence, line 190: A hand-written per-node config without allocation_fallback_strategy leaves node-less targets unassigned behind this one-time Info log. Would defaulting the fallback here, or logging at Warn, be safer?

config/crd/bases/cloudwatch.aws.amazon.com_amazoncloudwatchagents.yaml
```diff
@@ -6630,9 +6630,10 @@ spec:
                   allocationStrategy:
                     description: |-
                       AllocationStrategy determines which strategy the target allocator should use for allocation.
-                      The current option is consistent-hashing.
+                      The options are consistent-hashing and per-node.
```
movence, line 6633: The Go doc comment on this field in amazoncloudwatchagent_types.go still mentions only consistent-hashing, so the next make manifests will revert this text. Worth updating the comment too?

cmd/amazon-cloudwatch-agent-target-allocator/allocation/per_node.go
wenegiemepraise: we can default the fallback strategy 👍

Aakash-Dantre:
Checklist posted for the PR author to move into the PR description under a `## PR Checklist` heading.

## PR Checklist

- [ ] Commits are squashed into a logical, reviewable set (one commit for a single change)
- [x] Commits and PR description comply with Amazon internal guidelines
- [x] `make` passes locally — `go build ./...` clean, `make impi` PASS, `make checklicense` PASS, unit tests PASS for all target-allocator and manifests packages
- [ ] All GitHub Actions checks on the PR are passing
- [ ] Integration test evidence: link to a passing run, or state N/A with the reason
- [x] New or updated integration test coverage: aws/amazon-cloudwatch-agent-test#720 (`per_node_test.go`)
- [x] New functionality has unit tests; bug fixes have a reproducing test
- [x] Config translation changes include updated golden files — N/A, no translator changes
- [ ] Breaking or customer-visible changes are called out in the PR description

> Note for the description: besides adding the per-node strategy, this PR changes collector
> registration for the existing `consistent-hashing` strategy too — the initial List now skips pods
> with an empty `spec.NodeName`, `watch.Modified` is handled, and pods are dropped as soon as they
> carry a `DeletionTimestamp` rather than on `Deleted`. That increases target churn during rollouts
> for existing users and is currently undisclosed.

